PR #2, opened 2019-01-19
Wrap assignments in conditional into parentheses

krzysiek1507:
It's just safer to wrap assignments in conditionals into parentheses, so a developer won't mistake assignment operator with comparison operator.

https://github.com/rubocop-hq/ruby-style-guide#safe-assignment-in-condition


fxn:
Hey, thanks for the path, but I prefer the existing style.

PR #7, opened 2019-01-30
Allow to ignore paths with regular expression

bpohoriletz:
Right now we need to exclude all files in a particular directory or
files with particular name we will need to pass a full list into an
ignore method. In order to simplify this workflow we could add ability
to ignore paths based on regular expression that Dir.glob can recognize.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -156,7 +156,15 @@ module Zeitwerk
     # @param paths [<String, Pathname, <String, Pathname>>]
     # @return [void]
     def ignore(*paths)
-      mutex.synchronize { ignored.merge(expand_paths(paths)) }
+      paths.flatten.each do |path|
+        expanded_path = expand_paths(path)
+
+        if Dir[expanded_path.first].empty?
+          mutex.synchronize { ignored.merge(expanded_path) }
+        else
+          mutex.synchronize { ignored.merge(Dir[*expanded_path]) }
+        end
+      end
     end
 
     # Sets autoloads in the root namespace and preloads files, if any.
--- a/test/lib/zeitwerk/test_ignore.rb
+++ b/test/lib/zeitwerk/test_ignore.rb
@@ -68,20 +68,25 @@ class TestIgnore < LoaderTest
   test "ignored directories are not eager loaded" do
     files = [
       ["x.rb", "X = true"],
-      ["m/a.rb", "M::A = true"],
-      ["m/b.rb", "M::B = true"],
-      ["m/c.rb", "M::C = true"]
+      ["a/s/a.rb", "A::S::A = true"],
+      ["b/n/z.rb", "B::N::Z = true"],
+      ["b/n/x.h", "B::N::X = true"],
+      ["b/n/s/b.rb", "B::N::S::B = true"],
+      ["c/c.rb", "C::C = true"]
     ]
 
     with_files(files) do
       loader.push_dir(".")
-      loader.ignore("m")
+      loader.ignore("**/s", "*.h")
       loader.setup
       loader.eager_load
 
-      assert_equal 1, loader.autoloads.size
+      assert_equal 7, loader.autoloads.size
       assert ::X
-      assert_raises(NameError) { ::M }
+      assert C::C
+      assert B::N::Z
+      assert_raises(NameError) { A::S::A }
+      assert_raises(NameError) { B::N::X }
     end
   end
 


fxn:
Related, I just pushed this clarification in the README https://github.com/fxn/zeitwerk/commit/e79b96c1f5b0fd32429d556535fdce0f137d48bc.

bpohoriletz:
Thanks! For the default Rails app most likely it's not needed - I'd really want this to be possible because I've seen unconventional approaches to a file structure of a project influenced by DDD and Ember.js 
In Ember.js unit test file lives right next to a file with a definition of a tested unit. This approach is currently impossible because test files are within autoload/eager load folders so they will be loaded in production.
In a [DD Approach](https://bpohoriletz.github.io/2018/10/26/domain-driven-rails-architecture-pattern.html) the idea is to have separate test suite per bounded context, often it's a separate folder and a module. Independent test suite will have independent environment and it's own spec/ folder that will be within autoload path. This causes issues when Rails tries to load those files since they're incompatible.
When I read that Rails 6 will have a new default loader I saw it as a great opportunity to have the functionality I need. Not sure if it will be helpful for anyone else so I will totally understand if this PR won't get merged.

fxn:
That sounds interesting, didn't know about that use case.

Zeitwerk is intended for general Ruby usage, so I am interested on that use case and will be supported.

Could you show me one particular real layout with regular and test files living together below the root directories? Which pattern would you use to ignore them? With that I'll visualize one real case, and it might help also in documentation.

bpohoriletz:
Hello and thank you =)
I'll post few comments - one per each layout

The first one is influenced by [RFC](https://github.com/emberjs/rfcs/blob/master/text/0143-module-unification.md) from Ember.js

I this case unit test files live within `app/` folder [(an example)](https://github.com/bpohoriletz/bpohoriletz.github.io/tree/layouts_demo/samples/domain_driven_rails_architecture_pattern/app/contexts). I've also added `contexts/` folder but that is not required - subfolders could exist within default Rails folders (`models/`, `helpers/` etc. ). Depending on the test framework that is used name of the test file ends with `_spec.rb` or `_test.rb` - that's the convention that exists now.

In this case we would like to autoload all files within `app/` folder except for ones that end with `_spec.rb` or `_test.rb`

In terms of code something like `loader.ignore('**/*_spec.rb', **/*_test.rb)`

fxn:
Awesome, thanks for the samples!

bpohoriletz:
The Second one is influenced by a [Counterintuitive Rails pt. 1 - Ivan Nemytchenko](https://www.youtube.com/watch?v=KtD32fO_owU&t=1382s). But except for the separation of the UI and business logic I've also added ability to load only a part of the environment that is needed in the particular context [representations](https://github.com/bpohoriletz/bpohoriletz.github.io/tree/layouts_demo/samples/domain_driven_rails_architecture_pattern/representations) and [domain](https://github.com/bpohoriletz/bpohoriletz.github.io/tree/layouts_demo/samples/domain_driven_rails_architecture_pattern/domain)
Since each one has it's uniq setup for the test suite `representations/` and `domain/` can't be directly autoloaded. Moreover if in the future `Gemfile` becomes `gemfile.rb` it may cause additional issues.

In this case we would like to exclude `spec/` or `test/` subfolders entirely. 

In terms of code something like loader.ignore('domain/spec', representation/spec)

bpohoriletz:
The third one is built on top of the previous one. In case we'd like to introduce even more granular structure of the test suite we may have  `spec/`, `test/`, `gemfile.rb` deeply nested - for example one within each bounded context [(an example)](https://github.com/bpohoriletz/bpohoriletz.github.io/tree/layouts_demo/samples/domain_driven_rails_architecture_pattern/domain/contexts/billing)

In this case we'd like to exclude all subfolders named `spec/` or `test/`

In terms of code something like loader.ignore('domain/**/spec', representation/**/spec)

bpohoriletz:
The last piece - these tree layout aren't exclusive and can be used together in a single repo

fxn:
Thanks very much for your work in this PR.

I wanted to do some changes to be able to support glob pattern expansion on reload, have tests that reflect the use cases that justify this feature, and also document it. So I wrote https://github.com/fxn/zeitwerk/commit/803eac6db44e603828e51dccaed2688bb57f0c92 myself, more efficient than to try to transmit all I wanted here and go back and forth with details. Of course, the commit gives you deserved credit.

Thanks! ❤️

PR #10, opened 2019-02-13
Add Loader#do_preload_abspath to process single absolute paths

ricardotk002:
The logic for pre-loading a single path was spread in several methods. This patch tries to unify them into a single method that can be reused.

Feel free to ignore it if everything was done that way on purpose, I wasn't really sure if it made sense to factor this logic out.

PS. Your work on this loader is amazing, it makes it really easy to grasp such a core concept in Ruby.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -150,13 +150,7 @@ module Zeitwerk
       mutex.synchronize do
         expand_paths(paths).each do |abspath|
           preloads << abspath
-          if @setup
-            if ruby?(abspath)
-              do_preload_file(abspath)
-            elsif dir?(abspath)
-              do_preload_dir(abspath)
-            end
-          end
+          do_preload_abspath(abspath) if @setup
         end
       end
     end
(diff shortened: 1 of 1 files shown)


fxn:
Good catch!

Would you mind renaming the method to `do_preload_abspath`, and the parameter name to `abspath`? The library tries hard to use naming consistently and keep track of what is absolute to be able to easily follow the code and now at each point with which kind of path name you are dealing with.

This is motivated by... good naming practice, but also because this consistency is not present in `dependencies.rb` and that is something I wanted to deliberately address here.

Need to write code guidelines documenting this. By the way, `file` and `dir` are fine because they mean absolute file and directory names consistently. In that case I chose a short name because the library internally works only with absolute and the alternatives would have been too verbose.

Thanks for taking the time to contribute this ❤️.

ricardotk002:
Sure thing, the consistency argument is sound to me. Just updated the patch with your suggestions.

fxn:
Fantastic, thank you man!

PR #11, opened 2019-02-15
Allow to specify an alternative namespace in push_dir

casperisfine:
Hi, first thanks for Zeitwerk.

For context I'm currently experimenting with integrating Zeitwerk in our [Lita](https://github.com/litaio/lita) based chat bot so that we can benefit from an autoloader in development.

And in our chat bot, the constant to file path convention is mostly respected except that the top level namespace is implicit. e.g. `AppName::Foo::Bar` is located in `app/foo/bar.rb` because everything is under that namespace anyway.

I can perfectly move files around, it's really not the end of the world, but looking at Zeitwerk's source it seemed very easy to support implicit namespace, so I decided to suggest the feature.

If you feel like this is unnecessary added complexity that's fine.

@fxn what do you think?

cc @etiennebarrie @rafaelfranca

### DIFF
--- a/test/lib/zeitwerk/test_push_dir.rb
+++ b/test/lib/zeitwerk/test_push_dir.rb
@@ -4,13 +4,13 @@ require "pathname"
 class TesPushDir < LoaderTest
   test "accepts dirs as strings and stores their absolute paths" do
     loader.push_dir(".")
-    assert loader.root_dirs == { Dir.pwd => true }
+    assert loader.root_dirs == { Dir.pwd => Object }
     assert loader.dirs.include?(Dir.pwd)
   end
 
   test "accepts dirs as pathnames and stores their absolute paths" do
     loader.push_dir(Pathname.new("."))
-    assert loader.root_dirs == { Dir.pwd => true }
+    assert loader.root_dirs == { Dir.pwd => Object }
     assert loader.dirs.include?(Dir.pwd)
   end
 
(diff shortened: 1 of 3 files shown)


test/lib/zeitwerk/test_push_dir.rb
```diff
@@ -19,4 +19,14 @@ class TesPushDir < LoaderTest
     e = assert_raises(ArgumentError) { loader.push_dir(dir) }
     assert_equal "the root directory #{dir} does not exist", e.message
   end
+
+  module MyApp
+  end
+
+  test "an alternative namespace can be passed" do
+    files = [["app/x.rb", "TesPushDir::MyApp::X = true"]]
+    with_setup(files, dirs: [['app', namespace: MyApp]]) do
```
casperisfine: One likely footgun here is that if `MyApp` is reloadable, we'd hold onto a dereferenced constant.

fxn:
Hi @casperisfine! Thanks a lot for writing this patch, it is a good patch editing types and all!

In principle this is a flexibility point that could trigger expectatives for other similar flexibility points... I prefer to keep this aspect clear and simple.

I think of Zeitwerk looking at the future, and don't want to support existing project trees with non-supported structures. The only flexibility point that you have is the inflector. The customizable inflector is thought as an easy way to deal with acronyms and that kind of thing in a deterministic way controlled by your project, and could be technically possible to be abused saying that directory "foo" actually defines namespace `Bar`. But that is just a consequence of the inflection logic in the autoloader, not a deliberate feature per se.

So, no, would prefer to keep it the way it is, but thanks for the patch anyway. ❤️

casperisfine:
> The customizable inflector is thought as an easy way to deal with acronyms and that kind of thing in a deterministic way controlled by your project, and could be technically possible to be abused saying that directory "foo" actually defines namespace `Bar`.

Yeah it's not exactly equal as the inflector only allow inflection of a single element e.g. in my case I'd need to return `"Foo::Bar"`.

But yeah, I totally understand your point, this was a just in case PR.

Thanks for your time.

fxn:
> Yeah it's not exactly equal as the inflector only allow inflection of a single element e.g. in my case I'd need to return `"Foo::Bar"`.

Yeah, you could say that the parent directory of `app` is a root directory, use the inflector, and tell Zeitwerk to ignore anything else at that top level. But it is all too forced. Was not exactly proposing that as a workaround, would not be a good one. Better reestructure if you are willing to do it.

Issue #12, opened 2019-03-01
Performance issue in CI


n-rodriguez:
Hi there!

I don't know if it's the right place for this issue, but Rails bug tracker is very busy so I open it here :)

My issue : I've tested my Rails app with Rails 6.0.0.beta2 and tried the new Zeitgest loader, and noticed that my test suite duration is twice longer than before (in Gitlab CI env)  :/

The context : 

1. I use [grosser/parallel_tests](https://github.com/grosser/parallel_tests) to run the test suite with 6 cpus, and usually it takes 6/7mn to run.

2. Last week I introduced Zeitwerk in a bunch of gems I use ([simple_navigation_bootstrap](https://github.com/jbox-web/simple_navigation_bootstrap), [active_settings](https://github.com/jbox-web/active_settings)) and noticed that the tests were a bit longer but didn't dig into it.

3. Finally, I've set up Zeitwerk in 9 gems that I use in this app, and noticed that the test suite was really longer : it took 12/13mn to run.

4. Now I use Zeitwerk in the Rails app and the test suite take 30mn to run :cry: 

By reverting the Rails loader to the classic mode it takes 13mn to run.

I could remove Zeitwerk in the gems I use and run some tests and I'm pretty sure it will solve the issue.
But I don't think it's the right direction to go.

Maybe I should eager load everything in CI env?

What do you think?

Thank you!

n-rodriguez:
I put `Zeitwerk::Loader.eager_load_all` in an initializer with Rails classic loader, so all gems using Zeitwerk were eager loaded and the test suite took 6mn :tada:

n-rodriguez:
Okay! I think I got it :

* switch Rails loader to Zeitwerk mode
* add `Zeitwerk::Loader.eager_load_all` in `spec/rails_helper.rb`

This will eager load all gems that use Zeitwerk and the Rails application itself but only when doing `bundle exec rspec` or when using parallel_test.

Now the test suite duration is back to 6/7mn :tada:

fxn:
Eager loading does basically the same as autoloading, only it does it upfront, instead of on demand. In theory, total wall clock would be the same, but your app refutes that so let's see why is that.

[...]

PR #20, opened 2019-03-12
Add support for multiple loaders cooperating over the same namespace

casperisfine:
To begin, I understand that this use case has been deliberately rejected as mentioned in the README and in `ExplicitNamespace` comments. However I'd like to make a case for it.

### Why is it needed?

The Zeitwerk integration in Rails (that you wrote) use multiple loaders to handle different autoloading behavior. There's `Rails.autoloaders.main` for `config.eager_load_paths` and  `Rails.autoloaders.once` for `config.autoload_once_paths` (and I'm about to submit a PR to add two more).

Because of this the following scenario won't work:

```ruby
# config/application.rb
config.autoload_once_paths << root.join('lib')

# app/models/pricing.rb
class Pricing < AR::Base
end

# lib/pricing/calculator.rb
class Pricing::Calculator
end
```

What is happening here, is that `Rails.autoloader.once` will record `lib/pricing/` for it to be autovivified, but will never receive the `on_namespace_loaded` callback, hence won't register an `autoload` for `Calculator` after `Pricing` is loaded.

### Note on the implementation

Currently this patch calls `on_namespace_loaded` on all loaders when a directory is autoloaded, we could certainly optimize this by registering the loaders which need to be called.

But I'd rather validate the approach with you first.

cc @fxn @rafaelfranca @Edouard-chin

### DIFF
--- a/lib/zeitwerk/explicit_namespace.rb
+++ b/lib/zeitwerk/explicit_namespace.rb
@@ -32,7 +32,7 @@ module Zeitwerk
       # @return [void]
       def register(cpath, loader)
         mutex.synchronize do
-          cpaths[cpath] = loader
+          (cpaths[cpath] ||= []) << loader
           # We check enabled? because, looking at the C source code, enabling an
           # enabled tracer does not seem to be a simple no-op.
           tracer.enable unless tracer.enabled?
@@ -43,7 +43,12 @@ module Zeitwerk
       # @param loader [Zeitwerk::Loader]
       # @return [void]
       def unregister(loader)
-        cpaths.delete_if { |_cpath, l| l == loader }
+        cpaths.each do |cpath, loaders|
+          loaders.delete(loader)
+          if loaders.empty?
+            cpaths.delete(cpath)
+          end
+        end
         disable_tracer_if_unneeded
       end
 
@@ -59,8 +64,8 @@ module Zeitwerk
     @tracer = TracePoint.new(:class) do |event|
       # Note that it makes sense to compute the hash code unconditionally,
       # because the trace point is disabled if cpaths is empty.
-      if loader = cpaths.delete(event.self.name)
-        loader.on_namespace_loaded(event.self)
+      if loaders = cpaths.delete(event.self.name)
+        loaders.each { |l| l.on_namespace_loaded(event.self) }
         disable_tracer_if_unneeded
       end
     end
--- a/lib/zeitwerk/loader/callbacks.rb
+++ b/lib/zeitwerk/loader/callbacks.rb
@@ -23,7 +23,7 @@ module Zeitwerk::Loader::Callbacks
     log("module #{autovivified_module.name} autovivified from directory #{dir}") if logger
 
     loaded.add(autovivified_module.name)
-    on_namespace_loaded(autovivified_module)
+    Zeitwerk::Registry.loaders.each { |l| l.on_namespace_loaded(autovivified_module) }
   end
 
   # Invoked when a class or module is created or reopened, either from the
--- a/test/lib/zeitwerk/test_multiple.rb
+++ b/test/lib/zeitwerk/test_multiple.rb
@@ -54,4 +54,66 @@ class TestMultiple < LoaderTest
       assert App1::Foo::Bar::Baz
     end
   end
+
+  test "multiple loaders sharing an autovivified namespace" do
+    loaders = [loader, Zeitwerk::Loader.new]
+
+    files = [
+      ["lib0/namespace/foo.rb", <<-EOS],
+        module Namespace
+          class Foo
+          end
+        end
+      EOS
+      ["lib1/namespace/bar.rb", <<-EOS]
+        module Namespace
+          class Bar
+          end
+        end
+      EOS
+    ]
+    with_files(files) do
+      loaders[0].push_dir("lib0")
+      loaders[1].push_dir("lib1")
+      loaders.each(&:setup)
+
+      assert ::Namespace
+      assert ::Namespace::Foo
+      assert ::Namespace::Bar
+    end
+  end
+
+  test "multiple loaders sharing an explicit namespace" do
+    loaders = [loader, Zeitwerk::Loader.new]
+
+    files = [
+      ["lib0/namespace.rb", <<-EOS],
+        module Namespace
+        end
+      EOS
+      ["lib0/namespace/foo.rb", <<-EOS],
+        module Namespace
+          class Foo
+          end
+        end
+      EOS
+      ["lib1/namespace/bar.rb", <<-EOS]
+        module Namespace
+          class Bar
+          end
+        end
+      EOS
+    ]
+    with_files(files) do
+      loaders[0].push_dir("lib0")
+      loaders[0].tag = 'lib0'
+      loaders[1].push_dir("lib1")
+      loaders[1].tag = 'lib1'
+      loaders.each(&:setup)
+
+      assert ::Namespace
+      assert ::Namespace::Foo
+      assert ::Namespace::Bar
+    end
+  end
 end
--- a/test/lib/zeitwerk/test_unload.rb
+++ b/test/lib/zeitwerk/test_unload.rb
@@ -91,16 +91,16 @@ class TestUnload < LoaderTest
       la = Zeitwerk::Loader.new
       la.push_dir("a")
       la.setup
-      assert Zeitwerk::ExplicitNamespace.cpaths["M"] == la
+      assert_includes Zeitwerk::ExplicitNamespace.cpaths["M"], la
 
       lb = Zeitwerk::Loader.new
       lb.push_dir("b")
       lb.setup
-      assert Zeitwerk::ExplicitNamespace.cpaths["X"] == lb
+      assert_includes Zeitwerk::ExplicitNamespace.cpaths["X"], lb
 
       la.unload
-      assert_nil Zeitwerk::ExplicitNamespace.cpaths["M"]
-      assert Zeitwerk::ExplicitNamespace.cpaths["X"] == lb
+      refute Zeitwerk::ExplicitNamespace.cpaths.key?("M")
+      assert_includes Zeitwerk::ExplicitNamespace.cpaths["X"], lb
     end
   end
 end


fxn:
ACK!

Let me think about it. There is a tension in how to define the behaviour which is: Totally independent loaders whose existence I do not know vs different loaders that I control. You have to find your place there.

If you _know_ that other loaders are managing the same namespace, then you may need to load those dependencies first. That is a rule I apply to Ruby generally, regardless of this lib (and that influenced this choice). That is, if I am reopening `M`, which is a namespace owned by a different project, gem, loader, etc. then I must load `M` first in my project, so that everything flows.

In this view, someone is _in charge_ of defining `M`, and the rest are _users_ of `M`. The users of `M` cannot define `M` by any means, even less if `M` is a class (which superclass?), and thus they have to make sure they don't by preloading `M` (in the Ruby sense, not Zeitwerk sense).

For example, when I reopen something that belongs to a different project I either `require` the thing I am reopening, or user `module_eval` instead of the module keywork to be 100% the constant is present.

That was more or less my rationale for the way things work in the current version.

In constrast, for some things the implementation tries to have global awareness. That is why it uses global API like `autoload?` or `const_defined?`, to coordinate globally in some selected use cases.

But, usage and real use cases may influece this, I'll sleep on it a bit.

casperisfine:
Just to say that I perfectly agree with your assumptions, in the context where that other loader is a dependency.

But what broke those assumptions in my case is the current integration in Rails since it's very common in Rails app to have a single namespace in several load paths (`app/models`, `app/jobs`, etc).

A different take on this could be to rethink the current integration, and have `Zeitwerk::Loader#push_dir` take a configuration (`once: true/false`). But that would complicate `Loader` quite a bit, and I have to say I really like how simple the multi loader solution is.

fxn:
The reference to `app/models` and `app/jobs` makes me wonder. Let me make sure we are in sync with the current behavior:

1. In the directories managed by a loader you can have as many directories spreading a namespace as you want. So, `app/models/foo`, `app/jobs/foo`, `app/controllers/foo`, etc. will load the entire `Foo` namespace on demand, assuming that loader owns `Foo`.

2. On `setup`, if another loader finds a directory called `foo` and the constant `Foo` is defined, it recurses on the spot. That one is not lazy, if for whatever reason someone else defined the namespace, you go down and set autoloads right away.

In the case of a namespace existing in the directories managed by `once`, the owner of the namespace has to be `once`. Why? Because if you have `lib/foo/x.rb` defining `Foo::X` and the owner is `main`, on reload the original module stored in `Foo` will be gone, and `Foo::X` will become stale. Nobody is going to evaluate `lib/foo/x.rb` again because it belongs to `once`.

The `once` loader has to be the owner of `Foo`, and `main` the user. `Foo` won't be reloaded (which is correct), and the constants that belong to `Foo` managed by `main` (`app/models/foo/basr.rb`) will be autoloaded and reloaded cleanly because of the rule (2) above.

How does that sound?

byroot:
> How does that sound?

It's a bit late so I'll re-read all that tomorrow morning, but it makes sense right now.

If I understand you correctly, the logical conclusion is that we have to ensure that `once` has priority over `main` in Rails's Zeitwerk integration. Is that right?

So it would mean [inverting the two `yields` here](https://github.com/rails/rails/blob/49676e40547b8dcbd38aa927cfd7995957504493/railties/lib/rails/autoloaders.rb#L30-L31), so that `once` is always `setup` before `main` ? 

That being said I'm don't think we're describing the same issue. I wasn't exactly telling the (whole) truth when I was talking about `main` and `once`. So now I feel like I have to explain my ulterior motive. In the PR body I said:

> and I'm about to submit a PR to add two more

That's because we're having issues with integrating Zeitwerk in our app, mainly because we still have a few paths that are in `config.autoload_paths`, but not in `config.eager_load_paths`, meaning they are autoloaded, but never eager loaded (yes it's bad :/). So my ulterior motive here is add that old behavior to the Zeitwerk integration, and [right now I do it with even more loaders](https://github.com/rails/rails/compare/master...Shopify:zeitwerk-autoload-paths).

So in this instance I'm not even having reloadable ownership issues, because the "fight" wasn't between `main` and `once` but between `main` and `lazy`. I only transposed the description of the problem with `main` and `once` to make it easier to understand.

The issue arise before any namespace is loaded. The problem is that two loaders registers the same namespace, e.g.:

```
lib1/
  - foo/
    - bar.rb
lib2/
  - foo/
    - eggspam.rb
```

Assuming the `lib1` loader is setup before `lib2`, when `Foo` will be autovivified (or loaded is there is a `foo.rb`), only `lib1` will receive the `on_namespace_loaded`, callback because of how the registry works. It wouldn't have been a problem before https://github.com/fxn/zeitwerk/commit/8aa05dae5dda74c986e3c587906099203af53c1f because each loader had it's trace point, so they'd all receive the callback, and would all be able to set autoload for the `foo/bar.rb` and `foo/eggspam.rb` respectively.

But now that there is only a single trace point, `lib2` won't know `Foo` just got loaded, and won't ever set an autoload for `Foo::Eggspam -> foo/eggspam,rb`.

Let me know if what I'm saying makes sense.

fxn:
Hey, I have a busy day today. Let me just write a quick followup, and we'll continue.

The loaders in Rails have to be in the current order to respect precedence. Mimicks `require` and `$LOAD_PATH`. If you have `app/models/foo.rb` and `engine/app/models/foo.rb`, then Zeitwerk ignores the second one because setting up `once` is aware of the fact that there is already an autoload for `Foo` in a file. Same way as `require` only loads the first occurrence of a relative name.

What I meant was that if a gem that is not `activerecord` uses a top-level directory called `active_record`:

```
my_gem/lib/active_record
```

this is not forbidden, but then it is the responsibility of the gem to ensure `ActiveRecord` is not autoloaded from its directory. The gem does not own `ActiveRecord`, it has to make sure it _reopens_ `ActiveRecord`. Normally, that is accomplished with a `require "active_record"` somewhere at boot, and Zeitwerk knows it has to descend right away because the constant is present in memory.

Similarly, if you have a namespace shared in `once` and in `main`, the owner has to be `once`, because `once` is not reloaded. So boot has to issue a `require` call to load from `lib` early or something before `main` is setup.

I believe Zeitwerk cannot work with any possible combination of files and directories. I feel that it has to have constraints that make it sound, that make the problem solvable.

Currently, you know there are hard constraints about the project structure. If the structure is not complaint, you just cannot use Zeitwerk, and that is fine.

And it could be the case that I need to add constraints about namespace ownership too. But those I need to discover. By now, my intuition designing is that diferent loaders have to be responsible for different namespaces (think gems with their unique top-level namespace). Otherwise, if reopening, someone has the ownership, and the rest are users, and it cannot be the case that all are able to reload when they please.

I envision as design direction namespace ownership, and perhaps one single loader reloading, the rest only autoloading/eager loading.

That is intuition, but let's see what we learn in this discussion.

Regarding eager load vs autoload, in the early prototypes of Zeitwerk there was a method `do_not_eager_load` that I removed in https://github.com/fxn/zeitwerk/commit/a1bd83b10521f41f7f74e921602839a1813d11e4. Would that method help your use case and remove the need for additional loaders? Now, with the single trace point that we carefully enable and disable I think it could be back if there are use cases for it like you have.

casperisfine:
> I believe Zeitwerk cannot work with any possible combination of files and directories. I feel that it has to have constraints that make it sound, that make the problem solvable.

I totally agree with that. However as a general rule, if there are patterns it doesn't support, it should try to blow up early rather than fail to find constants that are "obviously there" to the user.

I'm fine with requiring some level of understanding to gem owners, if you write some Active Record plugin, it's not unrealistic to require from you to ensure AR is properly loaded etc.

However inside a regular Rails app, I believe the required level of understanding should be a bit lower.

I've setup the following repo to showcase something that will without a doubt happen to some users:  https://github.com/casperisfine/zeitwerk-test-bed

I think in such case it would be fine to raise an error early saying such case is not supported, but I'm afraid ignoring the directory like Zeitwerk does currently might cause a lot of confusion and support load. 


> Regarding eager load vs autoload, in the early prototypes of Zeitwerk there was a method do_not_eager_load that I removed in a1bd83b. Would that method help your use case and remove the need for additional loaders?

I believe so, and I think it's a much better way to go. I'll play with a branch with that feature added back and see how it goes. 

Overall I think the current separation between loaders make sense, but what creates all this confusion is the Rails integration using multiple loaders. 

Do you think it would make sense to add more capabilities to `Loader` so that `main` and `once` could be merged?

fxn:
I have looked at the sample code, but I am still unconvinced.

In general, most of the time gems and apps won't share namespaces. Nokogiri has its own namespace, and anyone else extending `Nokogiri` must _reopen_, therefore _preload_. This is true or good practice in Ruby itself, and since in Zeitwerk a file or directory is the same as writing a `require` conceptually, the same applies.

If a Rails application has `app/models/foo/a.rb` and `lib/foo/b.rb`, and wants to put `lib` in `autoload_once_paths` then we cannot share namespaces, because you are saying you do not want `lib/foo/b.rb` to be reloaded, but if you do not control who defines, who is owner of `Foo`, and leave it at `main`, then at some point `Foo` is going to be wiped, and you'd need to reload `lib/foo/b.rb`. But at the same you said you didn't want to.

There are here two solutions, I believe. One is to put `lib` in `autoload_paths`. With that solution all works out of the box. The other solution if you want to keep it in `autoload_once_paths` is to control the ownership of `Foo` by defining the `Foo` module early yourself using the `module` keyword at boot. In this option, `Foo` is not autoloaded, never removed, and both loaders can coexist.

Indeed, in the blog post I said

> as well as anything in the somewhat unknown config.autoload_once_paths (whose future is not bright)

I wrote (whose future is not bright) as a fun way to say I am very suspicious about the semantics of `autoload_once_paths` and my plan is to remove it in the future. People can always define their own loaders in applications if they have something very special the control, there is not need to stick to `main` and `once`.

matthewd:
ISTM maybe namespaces could/should be managed by zeitwerk globally, such that multiple loaders can own constants inside. (And it would keep track of who does, and only remove the namespace module when all loaders have unloaded themselves.)

If you're defining stuff inside a module that someone else has explicitly autoloaded, as opposed to an autovivified namespace, things might still need more thought... but in the simpler case where noone has claimed explicit ownership of the top level name, it feels pretty reasonable to keep it shared. The top-level `Object` namespace isn't first-come-first-served, and I think one can extend that to any other pure namespace. (That also gives a clear ownership rule that minimizes loaders needing to fight over ordering: ownership goes to the first one to contain a _file_ -- not directory -- matching the namespace in question, and it's correspondingly an immediate error if a second tries to claim it.)

Defining constants inside someone else's namespace really isn't _that_ unusual, whether from an application or another library, so I'm pretty dubious that "don't do that" is viable here.

fxn:
Let's think about it, I believe we are identifying a corner case that needs to be addressed, and it is a bit hard to see how.

It could be that we need to share global information related to this somehow. It could be that Rails should have 3 loaders, `main`, `once`, and `engines`, and perhaps the Rails integration is responsible for setting up `once` before `main`. It could also be that we leave it as is and address this in the documentation, establishing namespace rules like we have rules for project structure.

My vision for Zeitwerk is long-term, it has to work with normal Ruby projects, but it does not need to work with any Ruby project existing today. I have a degree of flexibility there. You are going to add Zeitwerk code to your gem, so a little adjustment could be fine if it makes sense, like writing an inflector is.

In general, my intuition gravitates towards _ownership_ and natural dependency loading that it is common in Ruby projects. In the sense that you load the dependencies that you depend on, even more if you are reopening something.

It is true that if you have a company-wide pure namespace called `Shopify` with no code (the equivalent of an autovivified module)... perhaps... it does not matter who loads first, that is true, and maybe there is something there.

That somehow connects to the another intuition which is that you can rely on _the past_, but not on _the future_. You have to think that there might be [a dozen loaders](https://github.com/fxn/zeitwerk/issues/12) responsible for their project trees, and you setup them generally serially. Things that concern to your loader need to be known _on setup_, and that generally just works because you loaded your dependencies first, again, especially if you are reopening something, or the documentation stresses that you have to.

A gem [could even ship with its code eager loaded](https://github.com/nanoc/nanoc/blob/d77313cc003308a57679811cc0130d30eed2eda2/nanoc-core/lib/nanoc/core.rb#L32), for example, so it is late for future loaders to change what that gem already assumed.

fxn:
> Defining constants inside someone else's namespace really isn't that unusual, whether from an application or another library, so I'm pretty dubious that "don't do that" is viable here.

@matthewd just to be clear, you can do that today. Given:

```
$ ls -R
active_record

./active_record:
foo.rb
$ cat active_record/foo.rb
module ActiveRecord::Foo; end
```

you can autoload and reload `Foo` within `ActiveRecord`:

```
$ irb -rzeitwerk -ractive_record
irb(main):001:0> l = Zeitwerk::Loader.new; 1
=> 1
irb(main):002:0> l.push_dir(".")
=> true
irb(main):003:0> l.setup
=> true
irb(main):004:0> ActiveRecord::Foo
=> ActiveRecord::Foo
irb(main):005:0> l.loaded
=> #<Set: {"ActiveRecord::Foo"}>
irb(main):006:0> l.reload
=> true
irb(main):007:0> l.loaded
=> #<Set: {}>
irb(main):008:0> ActiveRecord::Foo
=> ActiveRecord::Foo
```

fxn:
@casperisfine in case it helps, just wanted to let you know that in Rails `master` engines are loaded now by the `main` autoloader. See https://github.com/rails/rails/issues/35618#issuecomment-473216821.

fxn:
Usage may make me revise this, but as an initial resolution my hunch is to promote a clear understanding of who defines, and who reopens, and [have just written some rules of thumb that cover this](https://github.com/fxn/zeitwerk/commit/9e69b8efe7673bbcf479b7f4be5408ec8226afe4).

Thanks as always @casperisfine!

PR #22, opened 2019-03-14
Bring back the do_not_eager_load feature

casperisfine:
As discussed in https://github.com/fxn/zeitwerk/pull/20#issuecomment-472324682 I'm trying to bring back the possibility to have directories being autoloaded, but not eager loaded.

This is more or less a revert of https://github.com/fxn/zeitwerk/commit/a1bd83b10521f41f7f74e921602839a1813d11e4

I tested it with [this Rails patch](https://github.com/rails/rails/compare/master...Shopify:zeitwerk-autoload-paths-2) and it seems to work as expected.

@fxn would you be ok with bringing that feature back?

cc @rafaelfranca @Edouard-chin

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -89,6 +89,10 @@ module Zeitwerk
     attr_reader :lazy_subdirs
 
     # @private
+    # @return [Set<String>]
+    attr_reader :eager_load_exclusions
+
+    # @private
     # @return [Mutex]
     attr_reader :mutex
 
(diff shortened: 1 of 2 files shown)


fxn:
Sounds good!

If there is no sensitive data involved, would you mind sharing a real use case?

casperisfine:
> would you mind sharing a real use case?

Well, I can describe a bunch of cases in our app where it's used, but ultimately if we had unlimited manpower all of them could be refactored out very carefully, and we'd like to do it but it's a gigantic yak-shave. So the "real use case" mostly come down to "legacy™".

fxn:
Understood :).

Might need some time before working on this patch, can you work against this branch meanwhile to not block you?

casperisfine:
> can you work against this branch meanwhile to not block you?

Yes, no problem at all.

Also more generally about the usefulness of non eagerloaded paths and `autoload_once` paths you mentioned on the other PR: I'd agree with you about their very limited usefulness, and general "smell" status.

I'd totally be behind deprecating them, and starting to emit warning for them, but IMO 6.0 should still support them. Just my 0.02$.

fxn:
@casperisfine could you please point to `master`? If all works as expected I'll do a release.

Once published, I'll update the Rails integration to avoid eager loading anything in `autoload_paths` not present in `eager_load_paths` using this API. Are you doing that by hand now? Would that change in the integration simplify your branch with regard to those eager load exceptions?

casperisfine:
> Are you doing that by hand now?

I'm using this rails branch: https://github.com/rails/rails/compare/master...Shopify:zeitwerk-autoload-paths-2

> Would that change in the integration simplify your branch with regard to those eager load exceptions?

Yes I'm already working with that change since a couple days, and it's night and day.

PR #25, opened 2019-03-18
Add a test case for singleton classes in TracePoint, and fix next vs break

casperisfine:
Followup to: https://github.com/fxn/zeitwerk/pull/24

Turns out we can't use break here, otherwise we get a `LocalJumpError: break from proc-closure` (which makes me wonder if we could have a callback without a closure, which perf wise could be interesting, but that's another story).

The test case to exercise this code path is a bit complicated, as need extra unloaded namespaces to ensure the loader stays enabled.

This only test singleton classes, I tried to generate a case for regular anonymous classes with `Class#new` and `Module#new`, but without success so far. I don't know why TracePoint doesn't trigger in these cases.

@fxn


fxn:
Indeed, looks like in practice we won't get called for anonymous classes or modules. At least, I cannot imagine right now how.

I have removed that part of the callback to simplify the code, and added a Ruby compatibility test in case this changes in the future.

PR #26, opened 2019-03-19
Pass the parent cpath as second arguments for camelization

casperisfine:
### Why ?

I've encountered a few cases where the same path segment is camelized in different ways in different namespaces (I know...). e.g. `Foo::API` and `Bar::Api`.

It's bad, and ultimately I'd like to clean it up, but I think Zeitwerk should make it easy to handle.

Right now it's possible by matching the path, but it's not as convenient as matching the constant path because you have to handle both explicit namespace and autovivified namespaces (and for the later, does the abspath have a trailing slash or not?).

```ruby
def camelize(basename, abspath)
  return 'API' if abspath.end_with?('foo/api', 'foo/api/', 'foo/api.rb') 
end
```

IMHO it's a bit fragile.

### Proposition

By passing the `cpath` instead, it makes the comparison much simpler and less finicky, e.g

```ruby
def camelize(basename, cpath, abspath)
  return 'API' if cpath == 'MyNamespace::Bar'
end
```

### Questions

When the parent namespace is `Object` I'm passing `nil`, but I guess a case could be made to pass `"Object"` or `"::"` so that this argument is always a string, hence won't require users to check for nil before using string methods.

### Cons

- I implemented it in a backward incompatible way, because I figured Zeitwerk usage, and Inflector customization is likely still limited today. However if you'd like to keep backward compatibility I have a few ideas how to do it.

- As we've learned in another PR, `Module#name` isn't exactly cheap, I'd understand the argument that this would slow down all calls for a minor improvement of the ease of use of the API.

@fxn What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ The inflectors that ship with Zeitwerk are deterministic and simple. But you can
 # frozen_string_literal: true
 
 class MyInflector < Zeitwerk::Inflector
-  def camelize(basename, _abspath)
+  def camelize(basename, _cpath, _abspath)
     case basename
     when "api"
       "API"
@@ -297,7 +297,9 @@ end
 
 The first argument, `basename`, is a string with the basename of the file or directory to be inflected. In the case of a file, without extension. The inflector needs to return this basename inflected. Therefore, a simple constant name without colons.
 
-The second argument, `abspath`, is a string with the absolute path to the file or directory in case you need it to decide how to inflect the basename.
+The second argument, `cpath`, is a string with the fully qualified name of the parent constant, e.g. `"SomeNanespace::SubNamespace"`.
+
+The third argument, `abspath`, is a string with the absolute path to the file or directory in case you need it to decide how to inflect the basename.
 
 Then, assign the inflector:
 
@@ -478,7 +480,7 @@ I'd like to thank [@matthewd](https://github.com/matthewd) for the discussions w
 
 Also, would like to thank [@Shopify](https://github.com/Shopify), [@rafaelfranca](https://github.com/rafaelfranca), and [@dylanahsmith](https://github.com/dylanahsmith), for sharing [this PoC](https://github.com/Shopify/autoload_reloader). The technique Zeitwerk uses to support explicit namespaces was copied from that project.
 
-Finally, many thanks to [@schurig](https://github.com/schurig) for recording an [audio file](http://share.hashref.com/zeitwerk/zeitwerk_pronunciation.mp3) with the pronunciation of "Zeitwerk" in perfect German. 💯
+Finally, many thanks to [@schurig](https://github.com/schurig) for recording an [audio file](http://share.hashref.com/zeitwerk/zeitwerk_pronunciation.mp3) with the pronunciation of "Zeitwerk" in perfect German. :100:
 
 ## License
 
--- a/lib/zeitwerk/gem_inflector.rb
+++ b/lib/zeitwerk/gem_inflector.rb
@@ -10,9 +10,10 @@ module Zeitwerk
     end
 
     # @param basename [String]
+    # @param _cpath [String]
     # @param abspath [String]
     # @return [String]
-    def camelize(basename, abspath)
+    def camelize(basename, _cpath, abspath)
       (basename == "version" && abspath == @version_file) ? "VERSION" : super
     end
   end
--- a/lib/zeitwerk/inflector.rb
+++ b/lib/zeitwerk/inflector.rb
@@ -9,9 +9,10 @@ module Zeitwerk
     #   Zeitwerk::Inflector.camelize("api", ...)              # => "Api"
     #
     # @param basename [String]
+    # @param _cpath [String]
     # @param _abspath [String]
     # @return [String]
-    def camelize(basename, _abspath)
+    def camelize(basename, _cpath, _abspath)
       basename.split('_').map!(&:capitalize!).join
     end
   end
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -361,7 +361,8 @@ module Zeitwerk
     # @return [void]
     def set_autoloads_in_dir(dir, parent)
       each_abspath(dir) do |abspath|
-        cname = inflector.camelize(File.basename(abspath, ".rb"), abspath)
+        cpath = parent == Object ? nil : parent.name
+        cname = inflector.camelize(File.basename(abspath, ".rb"), cpath, abspath)
         if ruby?(abspath)
           autoload_file(parent, cname, abspath)
         elsif dir?(abspath)
--- a/test/lib/test_inflector.rb
+++ b/test/lib/test_inflector.rb
@@ -2,7 +2,7 @@ require "test_helper"
 
 class TestInflector < Minitest::Test
   def camelize(str)
-    Zeitwerk::Inflector.new.camelize(str, nil)
+    Zeitwerk::Inflector.new.camelize(str, nil, nil)
   end
 
   test "capitalizes the first letter" do


fxn:
Hey! [I have documented](https://github.com/fxn/zeitwerk/commit/008cb5b04ea3419646a216fa23f9af89381e8c81) directories passed to the inflector in any of the two arguments don't have trailing slashes.

That was a conscious choice, consequence of using file iterators, because otherwise you need to be defensive. But was not documented.

Regarding the proposal, I'll think about it.

casperisfine:
👍 

I'd totally understand if you feel this isn't worth it.

fxn:
Hey @casperisfine how does your inflector look like today? Do you use a custom inflector that delegates to AS `camelize` as well as having custom logic?

casperisfine:
There's no particular secret sauce so I think I can share it:

```ruby
      module ZeitwerkInflector
        class << self
          def special_cases
            @special_cases ||= begin
              lines = Rails.root.join('config/autoloading.txt').read.lines.map(&:strip).reject(&:empty?)
              lines.map { |c| [c.underscore.freeze, c.freeze] }.to_h.freeze
            end
          end

          def camelize(basename, abspath)
            return 'Api' if abspath.end_with?(
              '/controllers/api',
              '/controllers/concerns/api',
              '/checkouts/api',
              '/checkout_updaters/api',
              '/checkout_validators/api',
              '/shopify_payments/api',
              '/[REDACTED]/api.rb',
              '/responder/api.rb',
            )

            return 'Url' if abspath.end_with?(
              '/theme/settings/url.rb',
              '/online_store/theme_setting/url.rb',
            )
            return 'Html' if abspath.end_with?(
              '/theme/settings/html.rb',
              '/online_store/theme_setting/html.rb',
            )
            return 'CURRENCY' if abspath.end_with?('/billing/currency.rb')
            return '[REDACTED]' if abspath.end_with?('/eagerlib/[REDACTED]')

            special_cases[basename] || basename.camelize
          end
        end
```

### Observations

  - I haven't started cleaning them up, but ultimately I'd like to rename the constants which break consistency (`Html` vs `HTML`).
  - For some I have no clue how it will be possible, see the `CURRENCY` one for instance. I'm afraid this will forever be a special case. If the parent `cpath` was passed, my whitelist could contain `Billing::CURRENCY`, and the inflector be slightly smarter to handle nesting. With the abspath I simply can't.
  - I don't want to rely too much on `inflectors.acronyms` because it's a global app configuration and might have other repercussions which are likely to push you in a corner. 
  - Instead I have a list of ~150 constant names that are to be handled as special cases. e.g. `SomethingID`. I don't want to make `ID` and acronym because that would break tons of stuff named `SomethingElseId`, and I don't want to make `SomethingID` an acronym either because I use a custom rubocop Cop to statically maintain that list, and it wouldn't be able to statically analyze the list of AS acronyms.
  - Also something related to the inflector that caused us a bunch of problems is that it isn't used when eager loading. We do eager load code on CI and production, so we had issues with constants not being loadable in local being discovered after PRs are merged and shipped, hence why we went with a rubocop solution. It would be interesting to have a Zeitwerk API to be able to validate that all the eager loaded files would be able to be auto loaded.  


### Conclusion

So feel free to close this PR, it's not a blocker at all for, and might not even be the best way to improve the inflector API.

However I have to say the inflector is probably Zeitwerk weakpoint from my experience, and I'm even more concerned about the Rails implementation relying on `AS::camelize`, for newer apps it's probably fine, but legacy codebase might have a hard time making it work when Zeitwerk will become the default in 6.1.

Hope this helps, and let me know if some parts are unclear.

fxn:
> Also something related to the inflector that caused us a bunch of problems is that it isn't used when eager loading. We do eager load code on CI and production, so we had issues with constants not being loadable in local being discovered after PRs are merged and shipped, hence why we went with a rubocop solution. It would be interesting to have a Zeitwerk API to be able to validate that all the eager loaded files would be able to be auto loaded.

That is a good point. In order to be consistent we should eager load with `const_get`. I have it in the branch [auto-eager-load](https://github.com/fxn/zeitwerk/tree/auto-eager-load). Would you like to give it a try?

casperisfine:
Seems to work as expected:

```
/tmp/bundle/ruby/2.5.0/bundler/gems/zeitwerk-f09659db5edc/lib/zeitwerk/loader.rb:669:in `const_get': uninitialized constant [REDACTED]
```

😉 

I'll fix the issues it uncovered and run some profile to see wether the perf impact is significant or not.

fxn:
Fantastic!

casperisfine:
Early feedback: The error message is as user friendly as the classic autoloader one:

IIRC it used to say: `expected to find it in path/to/file.rb but wasn't found` or something like that. Could be worth decorating the LoadError.

fxn:
That branch has a [new log message](https://github.com/fxn/zeitwerk/commit/f09659db5edc4897dd6f27a8a5243a7acd2e37d4), but the problem is that classic mode controls when it does autoloads.

Perhaps we could log that to `$stderr` so that we print the message both in autoloading and eager loading.

By the way, working on this branch today found a very weird interaccion between autoloading and `$LOADED_FEATURES`, the suite was leaking `X`s and could not understand why. I have reported it [here](https://bugs.ruby-lang.org/issues/15790).

casperisfine:
> That branch has a [new log message](https://github.com/fxn/zeitwerk/commit/f09659db5edc4897dd6f27a8a5243a7acd2e37d4),

Yes I saw it, but logs are too verbose for them to be enabled most of the time.

When I'm talking about user friendlyness here, I'm thinking of junior / intermediate developers for whom a more detailed error message would go a long way.

And even for confirmed developers, the error is telling what constant it hasn't found, but not where it was looking, so I need to manually search for it.

> the problem is that classic mode controls when it does autoload.

Even if it's only active in eager load mode, it would go a long way.

> By the way, working on this branch today found a very weird interaction between autoloading and $LOADED_FEATURES,

👀

casperisfine:
Ok, so that branch caught a good dozen of autoloading issues. I can say I feel much more confident about enabling it in non eager loaded environment if that branch makes it to master.

Now that this is all fixed, I'm triggering a profile.

fxn:
🙌

casperisfine:
Boot time seems on par (there always a lot of variance so hard to know for sure, but it's really in the same 1-2% ballpark).

As for profiles, they show that method as a big part of boot as a caller, but as marginal in itself:

```
  Mode: cpu(1000)
  Samples: 76010 (28.00% miss rate)
  GC: 19255 (25.33%)
...
     28663  (37.7%)         353   (0.5%)     Zeitwerk::Loader#const_get_if_autoload
     


   Mode: wall(1000)
   Samples: 73794 (25.06% miss rate)
   GC: 21870 (29.64%)
...
      25351  (34.4%)         321   (0.4%)     Zeitwerk::Loader#const_get_if_autoload
```

`0.5%` isn't exactly nothing, but at the same time I don't see much potential for improvement given how simple it is.

So huge 👍 for this branch.

fxn:
Fantastic, thanks a lot for checking all this Jean, that gives me also confidence in moving forward.

I am going to log the error in `$stderr` and release.

fxn:
Just published 2.1.5, goo night man! and thanks again! ❤️

fxn:
@casperisfine let's study the `CURRENCY` use case.

My intention with the current `camelize` is to add some vinegar to project structures that depart too much from the conventions.

If the inflection is common in the whole project for the same basename, that is fine, expected and fully supported. If the inflection depends on the namespace... my hope is to discourage that by making the interface limited. With the absolute path people can still match on the path prefix, anchoring in `lib/`, or `app/models`, or `app/[^/]+/billing`, hoping that _some_ pattern prefix won't resasonably match below the root directories (you don't have and won't have `App::Models`).

Would any of that be enough in that use case?

casperisfine:
Sorry for the late reply, I was focuses on https://github.com/fxn/zeitwerk/pull/46

> My intention with the current camelize is to add some vinegar to project structures that depart too much from the conventions.

I can understand that, and I do want to eliminate as much corner cases as possible from our codebase. But even with this goal in mind I know that some of them are legit and will persist. So I do want a clean enough™ way to handle them.

> Would any of that be enough in that use case?

Today's API is "enough" for sure, I mean it does work. It's just much slower and dirtier (or acid since we're talking vinegar) than I'd like.

Ultimately want I want is primitives allowing me to lookup the exceptions in more or less `O(1)`, e.g. generate a big hash of my exception list and do a simple lookup.

But because I only have the absolute_path I can't, I need to match the end of the string with my possible exceptions. Having the either the ruby namespace, or the path relative to the load path would allow me to generate a proper hash with the exceptions list.

fxn:
> But because I only have the absolute_path I can't, I need to match the end of the string with my possible exceptions. Having the either the ruby namespace, or the path relative to the load path would allow me to generate a proper hash with the exceptions list.

Can you provide a before/after comparison with a concrete example? What you have to write today, and what you would write if you got the parent class or module? What is wrong with `end_with?` given that it has a similar performance to computing a hash code? Does the current implementation force you to have a long list of `end_with?` and there goes a O(n)?

casperisfine:
Sorry for the late reply, I'm a bit swamped these days.

> What is wrong with end_with? given that it has a similar performance to computing a hash code? Does the current implementation force you to have a long list of end_with? and there goes a O(n)?

Yes, that's my issue with it. As you can see in the snippet I posted previously, I already have 14 `end_with?` and just had to add another one today.

However to be perfectly honest, it's more of a "code elegancy" concern than a performance one. Our inflector is invisible on boot CPU/WALL profiles, so It's really not a performance argument.

> Can you provide a before/after comparison with a concrete example?

I mostly want to be able to specify the special cases in a declarative way, e.g:

```ruby
class Inflector < Zeitwerk::Inflector
  SPECIAL_CASES = {
    'Billing' => { 'currency' => 'CURRENCY' }
  }
  def camelize(basename, cpath, _abspath)
    SPECIAL_CASES.dig(cpath, basename) || super
  end
end
```

fxn:
> Sorry for the late reply, I'm a bit swamped these days.

No prob at all man! I am also at RailsConf with limited time, and anyway I can work on Zeitwerk only in my spare time. Let's give it the time it needs :).

I totally feel your pain regarding the declarative style. But at the same time my intuition tells me that to resist and restrict this API little bit. That gut feeling could be wrong, and it might turn out that I need to come back to this, but that is what I am leaning to with from a mix of the experience so far, and my vision for the future, how would Zeitwerk influence project structures in projects that adopt it. The transition phase to the future may have some rough edges.

One possible compromise in Shopify's inflector could be something like this (written off the top of my head):

```ruby
SPECIAL_CASES_WITH_SUFFIX = %{
  'api' => {
    %w(
      /controllers/api
      /controllers/concerns/api
      /checkouts/api
      /checkout_updaters/api
      /checkout_validators/api
      /shopify_payments/api
      /[REDACTED]/api.rb
      /responder/api.rb
    ) => 'Api'
  },
  'url' => {
    %w(
      /theme/settings/url.rb
      /online_store/theme_setting/url.rb
    ) => 'Url' },
  'html' => {
    %w(
      /theme/settings/html.rb
      /online_store/theme_setting/html.rb
    ) => 'Html'
  },
  'currency' => { %w(/billing/currency.rb) => 'CURRENCY' },
  '[redacted]' => { %w(/eagerlib/[REDACTED]) => '[REDACTED]' }
}
```

with that:

```ruby
def camelize(basename, abspath)
  if sc = SPECIAL_CASES_WITH_SUFFIX[basename]
    sc.each do |suffixes, inflection|
      suffixes.each do |suffix|
        return inflection if abspath.end_with?(suffix)
      end
    end
  end

  special_cases[basename] || basename.camelize
end
```

We short-circuit the lookup loop by basename first too.

Perhaps a tad less painful?

fxn:
Going to close here by now, even with such a huge code base the last version of inflector we came with doesn't seem too terrible (provided the API is restricted).

Let's see what is the experience when Rails 6 is out.

fxn:
I have seen use cases in which people need to inflect stuff in 3rd party engines.

The root problem there is that inflection rules are global in a Rails app and engine authors do not have control about how inflections are going to affect their code. This is just wrong, but it is something that by now we inherit in Rails from how things were done over the years.

The needed code matching paths from RubyGems is pushing this decision to the limit, I don't like that interface for this use case.

I think we could reconsider this in a backwards compatible way.

fxn:
Basically, I wanted to promote coherence in how things are inflected, think "projects should inflect things the same way", so this friction wanted to add a bit of vinegar (a price some legacy projects might have to pay).

But inflections and engines vs apps is no longer "your fault", to say it somehow, that is a design issue in Rails itself, out of the control of users.

fxn:
However, I cannot deny it feels wrong to me to bend this API to workaround something that should be improved in a different project.

The real fix that would go to the root of the matter would be to have autoloaders per engine, I believe.

fxn:
WIP in [this branch](https://github.com/fxn/zeitwerk/tree/camelize).

The idea is that the `initialize` methods of the existing classes for inflectors have an optional third argument `namespace_name`. That way, existing code passing two arguments works, and existing subclasses invoking `super(a, b)` or just `super` work too.

The news is that you can now implement `camelize` with arity 3, and in that case you'll get the namespace name.

Also, I guess it would make sense to augment the `inflect` interface so that the key can be scoped to a namespace, like:

```ruby
inflect(["sql_statement", "Parsers::SQL"] => "SQLStatement")
```

What do you think?

PR #28, opened 2019-03-20
fix reload when constant has been reopened

### LINKED ISSUE #27: open class for override in reload case is not supported
ruby: 2.6.2
rails: 6.0.0.beta3

I have a model  defined in my engine:
```ruby
# app/models/rails_org/member.rb
class Member < ApplicationRecord
end
```
I overrided it in my main project
```ruby
# app/models/member.rb
require 'rails_org/member'
class Member < ApplicationRecord
end
```
in reload case,  only delete main model defined path from $LOADED_FEATURES

Sorry for my poor english,  if need more information,  please point out.

qinmingyuan:
Fixes #27

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -236,6 +236,9 @@ module Zeitwerk
           $LOADED_FEATURES.delete(path) if ruby?(path)
         end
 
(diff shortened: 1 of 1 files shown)


fxn:
Closing here, see the rationale in #27. Thanks for the patch anyway!

PR #30, opened 2019-03-20
Improve expand_paths memory usage

krzysiek1507:
It uses 33% less memory when `paths` is not an array and 50% when `paths` is an array.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -554,7 +554,7 @@ module Zeitwerk
     # @param paths [<String, Pathname, <String, Pathname>>]
     # @return [<String>]
     def expand_paths(paths)
-      Array(paths).flatten.map { |path| File.expand_path(path) }
+      Array(paths).flatten.map! { |path| File.expand_path(path) }
     end
 
     # @param message [String]


lib/zeitwerk/loader.rb
```diff
@@ -554,7 +554,7 @@ def dir?(path)
     # @param paths [<String, Pathname, <String, Pathname>>]
     # @return [<String>]
     def expand_paths(paths)
-      Array(paths).flatten.map { |path| File.expand_path(path) }
+      Array(paths).flatten.map! { |path| File.expand_path(path) }
```
casperisfine: If you are to use `map!`, might as well go all the way and use `flatten!` as well.

krzysiek1507: Unfortunately, we can't.
1. `Array(object)` returns `object` when `object` is an array so `flatten!` would change `paths`
2. `flatten!` returns [nil if no modifications were made](https://ruby-doc.org/core-2.2.0/Array.html#method-i-flatten-21).

fxn:
🤘

lib/zeitwerk/loader.rb
casperisfine: 1: good point, even though for an internal method it could be acceptable.

2: it just mean you have to split the expression in two.

krzysiek1507: 1. I think `paths` can come from outside e.g. in `preload`.
2. If `paths` was an array then it would change it and there is no way to avoid it.

PR #32, opened 2019-03-25
Optimize path removal from $LOADED_FEATURES

casperisfine:
A small benchmark is better than a long description, so:

Before:

```ruby
[1] pry(main)> Benchmark.realtime { Rails.autoloaders.main.reload }
=> 32.67625700007193
[2] pry(main)> $LOADED_FEATURES.size
=> 7413
```

After:

```ruby
[1] pry(main)> Benchmark.realtime { Rails.autoloaders.main.reload }
=> 0.8327639999333769
[2] pry(main)> $LOADED_FEATURES.size
=> 7413
```

### Why?

The previous implementation was basically `O(n*m)` with `n = $LOAD_PATH.size` and `m = autoloads.size`.

The new one is more or less `O(n)`.

### However

That being said, a profile showed that a good part of the time spent clearing `$LOADED_FEATURES` was spent in `bootsnap`:

```
  Mode: wall(1000)
  Samples: 13109 (67.42% miss rate)
  GC: 4846 (36.97%)
     12824  (97.8%)        4276  (32.6%)     Bootsnap::LoadPathCache::LoadedFeaturesIndex#purge
      6503  (49.6%)        2227  (17.0%)     #<Array:0x00007fc8218692f0>.delete
       695   (5.3%)         503   (3.8%)     Zeitwerk::Loader#set_autoload
     20102 (153.3%)         188   (1.4%)     Zeitwerk::Loader#unload
```

By using `reject!` we're bypassing `bootsnap` hook, so once I add support for `reject!` in `bootsnap`, it might be a bit slower than it is in this benchmark.

@fxn @Edouard-chin @rafaelfranca @burke

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -227,6 +227,9 @@ module Zeitwerk
     # @return [void]
     def unload
       mutex.synchronize do
+
+        unloaded_paths = Set.new
+
         autoloads.each do |path, (parent, cname)|
           if parent.autoload?(cname)
             parent.send(:remove_const, cname)
@@ -236,10 +239,14 @@ module Zeitwerk
             log("#{cpath(parent, cname)} unloaded") if logger
           end
 
-          # Let Kernel#require load the same path later again by removing it
-          # from $LOADED_FEATURES. We check the extension to avoid unnecessary
-          # array lookups, since directories are not stored in $LOADED_FEATURES.
-          $LOADED_FEATURES.delete(path) if ruby?(path)
+          unloaded_paths << path if ruby?(path)
+        end
+
+        # Let Kernel#require load the same paths later again by removing them
+        # from $LOADED_FEATURES. We check the extension to avoid unnecessary
+        # array lookups, since directories are not stored in $LOADED_FEATURES.
+        unless unloaded_paths.empty?
+          $LOADED_FEATURES.reject! { |path| unloaded_paths.include?(path) }
         end
 
         autoloads.clear


fxn:
Fantastic!

If we bypass `bootsnap`, Rails apps won't reload properly. Should we perhaps patch `bootsnap` first, then merge here, and bump their dependencies in Rails? Otherwise I could not release what's in `master` freely.

casperisfine:
> Should we perhaps patch `bootsnap` first, then merge here, and bump their dependencies in Rails?

Yeah I'm on it.

burke:
This makes sense, `$LOADED_FEATURES`'s change observer does a pretty expensive operation on each change. We could batch it, or defer it, or any number of things :+1:

casperisfine:
@fxn I just released https://github.com/Shopify/bootsnap/pull/252 as part of `bootsnap 1.4.2`.

fxn:
Thanks a lot for coordinating this @casperisfine.

I am doing some minor tweaks, one of them is to not store files in the set if the autoload is still configured but not executed. (There is [a compatibility test](https://stuart-team.atlassian.net/wiki/spaces/EN/pages/640057487/Full+DB+production+copy+for+development+in+Dev+environment) that checks `remove_const` does not trigger the autoload, since all this is undocumented.)

I'll soon merge, publish, and update Rails `master` too. Will write back.

fxn:
I have published version 1.4.3 of Zeitwerk with this patch.

The `master` branch of Rails has these updates too (https://github.com/rails/rails/commit/379d7e8bfa9484d495c2051266852df908c0ea59). (Not sure why did the `-java` version disappear from `Gemfile.lock`.)

burke:
Ah, we didn't have `bootsnap-1.4.2-java.gem` pushed; I just did that now.

fxn:
`Gemfile.lock` updated in Rails, thanks Burke!

casperisfine:
> Ah, we didn't have `bootsnap-1.4.2-java.gem` pushed;

My bad, I used the Shipit stack, maybe it need some update?

fxn:
Have documented in https://github.com/fxn/zeitwerk/commit/c4730992d7303a7d72df6c25beefc8d70113c2cb that we have to take that API into account when mutating `$LOADED_FEATURES`.

PR #34, opened 2019-04-03
Reduce memory retention by deduplicating strings

casperisfine:
### Context

We're currently having troubles shipping Zeitwerk in production because it's causing increased memory usage compared to the classic autoloader.

Memory problems are hard to diagnose, I tried many tools to narrow it down without much success. The most insightful info I have is that the majority of the difference is caused by `T_STRING` (~25% more than classic autoloader).

So I highly suspect the various data structures inside `Zeitwerk::Loader` to be responsible:

```
irb(main):008:0> Zeitwerk::Registry.loaders.first.autoloads.size
=> 7943
irb(main):012:0> Zeitwerk::Registry.loaders.first.shadowed_files.size
=> 2417
irb(main):016:0> Zeitwerk::Registry.loaders.first.loaded_cpaths.size
=> 7862
irb(main):018:0> ActiveSupport::Dependencies.loaded.size
=> 2390
```

(That last one is because some constants are loaded before Zeitwerk kicks in, we're in the process of cleaning that up).

If we compare to the classic autoloader:

```
irb(main):001:0> ActiveSupport::Dependencies.loaded.size
=> 9899
```

### Solutions

This PR is only one of the possible solutions that I want to explore. Ultimately I'd like to release paths references as much as possible in eager loading scenarios, but it's a bit tricky.

In the meantime I'm trying to reduce the impact of keeping all these references

### This PR

Here I'm trying to deduplicate the strings Zeitwerk holds onto. It's aimed at two types:

First I intern `cpath` strings since they are held inside `autoloads`, `loaded_cpaths`, `shadowed_files` and `lazy_subdirs`, so they can be duplicated up to 4 times.

Then, the second part is trickier. The main idea is to intern the paths that will be passed onto `Kernel#require`, so that the entry added into `$LOADED_FEATURES` isn't a duplicate. However it's a bit finicky because `Kernel#require` will call a bunch of FS functions on it like `realpath` etc, and most of these will return a tainted string with ASCII encoding. Tainted strings can't be interned, so we first need to untaint them. Then for the interning to work, the encodings have to match, so we call `force_encoding` as well.

### Considerations

I totally understand that this kinda complexifies the codebase, but IMHO the tradeoff is worth it.

### Further work

It's a bit more complex so I haven't started yet, but similar to what I proposed for the classic autoloader in https://github.com/rails/rails/pull/35747, I think Zeitwerk should try not to hold on all that data when we know it won't be reloading.

I'm unsure how this can be detected, `AS::Dependencies` have a clear `mechanism == :require`, Zeitwerk not so much.

@fxn @rafaelfranca @Edouard-Chin

### DIFF
--- a/lib/zeitwerk.rb
+++ b/lib/zeitwerk.rb
@@ -1,6 +1,27 @@
 # frozen_string_literal: true
 
 module Zeitwerk
+  class << self
+    # Kernel#require will apply String#-@ on paths before registering them in $LOADED_FEATURES.
+    # That method tries to intern the string to avoid holding onto multiple copies of it.
+    # However there is a couple problem that prevent the deduplication from happening in the vast
+    # majority of cases.
+    #
+    # First tainted string can't be interned, and `File.expand_path` do return tained string.
+    #
+    # Secondly, `Kernel#require` rebuilds a string from what is returned by the file system,
+    # and most of the time that string encoding isn't the application default encoding.
+    PATHS_ENCODING = $LOADED_FEATURES.first.encoding
+    def intern_path!(path)
+      -(path.untaint.force_encoding(PATHS_ENCODING))
+    end
+
+    CPATHS_ENCODING = Object.name.encoding
+    def intern_cname!(cpath)
+      -(cpath.dup.untaint.force_encoding(CPATHS_ENCODING))
+    end
+  end
+
   require_relative "zeitwerk/loader"
   require_relative "zeitwerk/registry"
   require_relative "zeitwerk/explicit_namespace"
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -198,7 +198,9 @@ module Zeitwerk
     def expand_ignored_glob_patterns
       # Note that Dir.glob works with regular file names just fine. That is,
       # glob patterns technically need no wildcards.
-      ignored_paths.replace(ignored.flat_map { |path| Dir.glob(path) })
+      paths = ignored.flat_map { |path| Dir.glob(path) }
+      paths.map! { |p| Zeitwerk.intern_path!(p) }
+      ignored_paths.replace(paths)
     end
 
     # Sets autoloads in the root namespace and preloads files, if any.
@@ -459,7 +461,8 @@ module Zeitwerk
       # $LOADED_FEATURES stores real paths since Ruby 2.4.4. We set and save the
       # real path to be able to delete it from $LOADED_FEATURES on unload, and to
       # be able to do a lookup later in Kernel#require for manual require calls.
-      realpath = File.realpath(abspath)
+      realpath = Zeitwerk.intern_path!(File.realpath(abspath))
+
       parent.autoload(cname, realpath)
       if logger
         if ruby?(realpath)
@@ -469,7 +472,7 @@ module Zeitwerk
         end
       end
 
-      autoloads[realpath] = [parent, cname]
+      autoloads[realpath] = [parent, Zeitwerk.intern_cname!(cname)]
       Registry.register_autoload(self, realpath)
 
       # See why in the documentation of Zeitwerk::Registry.inceptions.
@@ -554,7 +557,7 @@ module Zeitwerk
     # @param cname [String]
     # @return [String]
     def cpath(parent, cname)
-      parent.equal?(Object) ? cname : "#{parent.name}::#{cname}"
+      Zeitwerk.intern_cname!(parent.equal?(Object) ? cname : "#{parent.name}::#{cname}")
     end
 
     # @param dir [String]
@@ -563,7 +566,7 @@ module Zeitwerk
     def each_abspath(dir)
       Dir.foreach(dir) do |entry|
         next if entry.start_with?(".")
-        abspath = File.join(dir, entry)
+        abspath = Zeitwerk.intern_path!(File.join(dir, entry))
         yield abspath unless ignored_paths.member?(abspath)
       end
     end
@@ -583,7 +586,7 @@ module Zeitwerk
     # @param paths [<String, Pathname, <String, Pathname>>]
     # @return [<String>]
     def expand_paths(paths)
-      Array(paths).flatten.map! { |path| File.expand_path(path) }
+      Array(paths).flatten.map! { |path| Zeitwerk.intern_path!(File.expand_path(path)) }
     end
 
     # @param message [String]
--- a/lib/zeitwerk/loader/callbacks.rb
+++ b/lib/zeitwerk/loader/callbacks.rb
@@ -36,7 +36,7 @@ module Zeitwerk::Loader::Callbacks
       autovivified_module = parent.const_set(cname, Module.new)
       log("module #{autovivified_module.name} autovivified from directory #{dir}") if logger
 
-      loaded_cpaths.add(autovivified_module.name)
+      loaded_cpaths.add(cpath(parent, cname))
       on_namespace_loaded(autovivified_module)
     end
   end
--- a/test/lib/zeitwerk/test_explicit_namespace.rb
+++ b/test/lib/zeitwerk/test_explicit_namespace.rb
@@ -60,6 +60,42 @@ class TestExplicitNamespace < LoaderTest
     end
   end
 
+  test "the autoload path is re-used by Kernel#require" do
+    files = [
+      ["app/models/hotel.rb", "class Hotel; end"],
+    ]
+    with_setup(files, dirs: "app/models") do
+      abspath = loader.autoloads.keys.first
+
+      assert_equal File.expand_path("app/models/hotel.rb"), abspath
+
+      assert_kind_of Class, Hotel
+
+      assert_equal abspath, $LOADED_FEATURES.last
+      assert_same abspath, $LOADED_FEATURES.last
+    end
+  end
+
+  test "the various cpaths hold onto by the loader are deduplicated" do
+    files = [
+      ["app/models/hotel.rb", "class Hotel; end"],
+    ]
+    with_setup(files, dirs: "app/models") do
+      hotel_cname = Zeitwerk.intern_cname!('Hotel')
+      assert_equal 'Hotel', hotel_cname
+
+      autoload_cpath = loader.autoloads.values.first.last
+      assert_equal hotel_cname, autoload_cpath
+      assert_same hotel_cname, autoload_cpath
+
+      assert_kind_of Class, Hotel
+
+      loaded_cpath = loader.loaded_cpaths.first
+      assert_equal hotel_cname, loaded_cpath
+      assert_same hotel_cname, loaded_cpath
+    end
+  end
+
   # As of this writing, a tracer on the :class event does not seem to have any
   # performance penalty in an ordinary code base. But I prefer to precisely
   # control that we use a tracer only if needed in case this issue
--- a/test/lib/zeitwerk/test_ruby_compatibility.rb
+++ b/test/lib/zeitwerk/test_ruby_compatibility.rb
@@ -66,6 +66,49 @@ class TestRubyCompatibility < LoaderTest
     end
   end
 
+  test "require duplicates the passed path in $LOADED_FEATURES unless specific conditions are met" do
+    on_teardown do
+      delete_loaded_feature('users_controller.rb')
+    end
+
+    files = ["admin/users_controller.rb", "class UsersController; end"]
+    with_files(files) do
+      absolute_path = File.expand_path("admin/users_controller.rb")
+      require absolute_path
+      assert_equal absolute_path, $LOADED_FEATURES.last
+      refute_same absolute_path, $LOADED_FEATURES.last
+    end
+  end
+
+  test "Zeitwerk.intern_path! prevents require from duplicating the passed path in $LOADED_FEATURES" do
+    on_teardown do
+      delete_loaded_feature('users_controller.rb')
+    end
+
+    files = ["admin/users_controller.rb", "class UsersController; end"]
+    with_files(files) do
+      absolute_path = Zeitwerk.intern_path!(File.expand_path("admin/users_controller.rb"))
+      require absolute_path
+      assert_equal absolute_path, $LOADED_FEATURES.last
+      assert_same absolute_path, $LOADED_FEATURES.last
+    end
+  end
+
+  test "Zeitwerk.intern_path! works even with very exotic paths" do
+    on_teardown do
+      delete_loaded_feature('users_controller.rb')
+    end
+
+    path = "you-owe-me-a-🍺.rb"
+
+    files = [path, "class UsersController; end"]
+    with_files(files) do
+      absolute_path = Zeitwerk.intern_path!(File.expand_path(path))
+      require absolute_path
+      assert_equal absolute_path, $LOADED_FEATURES.last
+      assert_same absolute_path, $LOADED_FEATURES.last
+    end
+  end
   # While unloading constants we leverage this property to avoid lookups in
   # $LOADED_FEATURES for strings that we know are not going to be there.
   test "directories are not included in $LOADED_FEATURES" do


fxn:
Let me think about it.

Right now in theory you could unload after eager loading, I had in mind the possibility to forbid that, and on eager load just clear all these structures.

The trade-off there would be that you loose the opportunity of inspecting the loader state if there is something to debug, but that could be solved with a flag or something, since inspecting after eager loading should be exceptional. Exceptional use cases => exceptional interface.

fxn:
Ah, one exception would be `do_not_eager_load`, you might need to keep data to autoload. Just thinking out loud, need to reflect about it, but certainly this was something I wanted to work on somehow.

casperisfine:
> The trade-off there would be that you loose the opportunity of inspecting the loader state if there is something to debug

I agree that these have proven very useful for debugging, but as you say, one flag would be enough.

> Ah, one exception would be do_not_eager_load, you might need to keep data to autoload.

Yeah I though about that as well, that's mainly what makes it tricky. Sorry for making you restore it /o\

fxn:
I believe it is doable even with those exceptions. Like, if there are no exceptions just clear everything, and if there are, be a bit more careful, but trim it down to a minimum. Have not opened the editor, but from memory I believe it can be done.

I have a talk to give tomorrow and need to focus on it, but will tackle this later.

fxn:
Jean, do you use introspection API while eager loading.

I have not written anything yet, but the idea I have in mind is to document the inspection API in a debugging section. Under such approach, it feels right to document that by default none of that is available while or after eager loading.

Nowadays we delete from the lazy dirs collection. We could delete also from the autoloads one in the callbacks if impliding (working name) and avoid pushing anything to loaded paths. On my phone, just from my head.

If I am not mistaken, that approach would have a smaller footprint even while eager loading (vs clearing afterwards) and `do_not_eager_load` would not need special logic. The code by itself would leave the minimum leftovers.

I have delayed to document the introspection API and contract re `Kernel#require` on purpose, these need some real world usage and thought before committing to them so I can do that.

casperisfine:
> Jean, do you use introspection API while eager loading?

I did use it a few times when trying to figure out why some constant was shadowed etc. 

But the most useful debug tool by far was the logger, and I often edited Zeitwerk to log even more things.

fxn:
> I did use it a few times when trying to figure out why some constant was shadowed etc.

Perfect, such use case could be covered with the flag.

> But the most useful debug tool by far was the logger, and I often edited Zeitwerk to log even more things.

That is great, and also matches my experience.

fxn:
I have been thinking about this and have connected some previous thoughts.

For me, Zeitwerk is a fundamentally a _code loader_. Auto/Eager let's you choose _when_ you wanna load: on demand, or upfront. In my mind, reloading has always been in a different category, in a way. And it is kinda orthogonal: you could want to reload regardless of how you loaded. In Rails, that maps to the possible matrix of values for `config.cache_classes` and `config.eager_load`.

Regular gems that do not implement services are a first class user of Zeitwerk, and reloading does not make sense for them in general. You edit the code, run the test suite.

So, for me, the main use case of reloading is _services in development mode_. That is a very concrete use case seen from that perspective.

So I believe I am going to reflect all these thoughts in the interface: In order to reload, you are going to need to opt-in. The very interface is telling you reloading has a very specific use case, which is now just commented in passing in the documentation. In Rails that maps to `config.cache_classes`.

Of course, the Rails integration would hide all these for you, your interface is still `config.*`.

With that, the `autoloads` hash table, which is the one that keeps more metadata, will be kept to a minimum.

That would be a first step in RAM reduction, and I bet an important one. With that in place, we can see further tweaks are worthwhile.

(All this, modulus I have not written a single line of code 😄 .)

casperisfine:
> In order to reload, you are going to need to opt-in. 

:+1: to that.

fxn:
I think I have it, the work so far is in the [`reload`](https://github.com/fxn/zeitwerk/tree/reload) branch.

In order to reload, you need to opt-in before `setup`:

```ruby
loader.enable_reloading
loader.setup
```

I have implemented two chages that should reduce memory footprint notably.

Zeitwerk maintains two hashes with the paths to files and directories for which it sets autoloads. One in each loader instance to know which module and cname a given path corresponds to. And the other one in the global registry to be able to identify our own autoloads in the wrapped `Kernel#require`. So, if you are autoloading, those hashes only keep what has been visited because Zeitwerk is lazy. If you are eager loading, though, eventually you fill the two of them with what you have in the file sytem.

OK, I have implemented two optimizations:

No matter whether you reload or not, once a file is autoloaded, its entry is removed from the global registry.

Now, if you are not reloading, both hashes shrink as you autoload or eager load files, keeping them to a minimum. Similarly to what we already do with `lazy_subdirs`, for example, the entries are deleted as used. We can do that now because once autoloaded, the entries in each loader instance are used to know what to unload if reloading, but have no other utility if not.

If a project depends on 15 gems using Zeitwerk, this optimization compounds, since those gems generally won't have reloading enabled.

This approach by itself, without doing anything special, supports `do_not_eager_load`.

I'll let it rest for a while, and if all is good (mod tweaks) I'll publish 2.0.0 and update the Rails integration.

fxn:
Ah, let me add that when eager loading with reloading disabled these hashes expand and shrink on demand. It is not that they get full and then shrink. As autoloads are set in one given depth level, they expand a little, but right after it the files are required, so they shrink, and if there are namespaces some more autoloads are registered, thus expanding a bit again.

It is a dynamic balance. They become collections that expand and shrink as you walk the project and that have the minimum metadata at any given time, eventually ending up being voided if there are no eager load exceptions.

I like very much how it all flows and is more optimal by itself.

(Jean I believe you know all this, explaining details here mostly for the archives.)

fxn:
I have published 2.0.0, and `rails/master` depends on it (https://github.com/rails/rails/commit/57c7cbb1623c0e8befc58988a34bbb9896fd226a).

Let's see how's memory consumption now!

casperisfine:
Awesome. That's a lot Xavier.

I'll try to test this ASAP, might take a couple days because it mean we need to upgrade Rails. I'll keep you updated.

fxn:
Fantastic!

It is a pity that we have `loaded_cpaths`. I added that during the Rails integration because you can ask AS::Dependencies for that collection, but it was not written in the first versions of the gem, because as you said I expect less needs for introspections, and more use cases for troubleshooting, covered by logging.

I have my axe raised for `loaded_cpaths`, if I can come with a way to get rid of it in Rails, I'll probably remove it too.

fxn:
Hey! I think I have a plan for `loaded_cpaths`, if I am not mistaken we'll be able to have it empty if not reloading. If you can hold that memory test for a while, I'll see if my idea works out and write back soon.

casperisfine:
> If you can hold that memory test for a while

Too late:

Before:

```
retained memory by gem
   4.81 MB  zeitwerk-1.4.3

retained objects by gem
    51064  zeitwerk-1.4.3
```

After:
```
retained objects by gem
    19987  zeitwerk-2.0.0

retained memory by gem
   2.74 MB  zeitwerk-2.0.0
```

That being said these numbers are not to be taken as gospel, because these profiles already showed Zeitwerk using less memory than `AS::Dependencies`, but still ended consuming more in actual production.

That being said we very recently stripped over 150MB from Active Record, so I think we have enough leeway now to ship Zeitwerk in prod even if it's using marginally more memory than `AS::D`. If you improve it further soon it's just cherry on top.

fxn:
Hey! Those tests are good!

> That being said we very recently stripped over 150MB from Active Record

Wow, how?

> I think we have enough leeway now to ship Zeitwerk in prod

I'll open some champagne when that happens :).

Regarding my idea to optimize RAM even more, I need to sleep on it, but I have [this branch (WIP)](https://github.com/fxn/zeitwerk/tree/loaded_cpaths) where we shrink `autoloads` _always_, and move the entry to `to_unload` if reloading is enabled. That will save memory both in development and in production.

In the Rails code base, the [descendants tracker](https://github.com/rails/rails/blob/a497ece3f7046123cb7fe6720f08d40e4ca6019d/activesupport/lib/active_support/descendants_tracker.rb) is the only use case that I know that needs that information. See the call to `ActiveSupport::Dependencies.autoloaded?` in `clear`:

```ruby
def clear
  if defined? ActiveSupport::Dependencies
    @@direct_descendants.each do |klass, descendants|
      if ActiveSupport::Dependencies.autoloaded?(klass)
        @@direct_descendants.delete(klass)
      else
        descendants.reject! { |v| ActiveSupport::Dependencies.autoloaded?(v) }
      end
    end
  else
    @@direct_descendants.clear
  end
end
```

but that code does not really need to know if the thing was autoloaded, what it needs to know is _if it would be unloaded during a reload_. So, we only need to provide `to_unload?`, and that predicate is always false if reloading is disabled. Also, that predicate is wrong, it receives an object, but you cannot know if an object was autoloaded (think `X = 1`), what you can know is if _a cpath_ was autoloaded.

All in all, with this observation we can refine slighltly the interface and do not need to store the autoloaded cpaths if reloading is disabled.

I'll continue working in that direction. Have a trip to Italy on Wednesday, but unless I find something unexpected, I'll try to have it ready for Rails RC1.

casperisfine:
> Wow, how?

Schema cache deduplication: https://github.com/rails/rails/pull/35891, plus various other string interning there and there. 

>  So, we only need to provide to_unload?, and that predicate is always false if reloading is disabled.

Totally makes sense. This striping could be quite noticeable. Let me know if you need to shoot ideas or whatever else I can help with.

fxn:
2.1.0 published with that last optimization, and `rails/master` updated too.

casperisfine:
Interesting, I just ran the profile again, and 2.1 seem to use a bit more memory than 2.0:

```
retained memory by gem
   3.39 MB  zeitwerk-2.1.0

retained objects by gem
    26269  zeitwerk-2.1.0

retained objects by location
     7435  /tmp/bundle/ruby/2.5.0/gems/zeitwerk-2.1.0/lib/zeitwerk/loader.rb:610

retained memory by location
    956.37 kB  /tmp/bundle/ruby/2.5.0/gems/zeitwerk-2.1.0/lib/zeitwerk/loader.rb:515
```

fxn:
@casperisfine how strange... which mode is that? development or production?

casperisfine:
@fxn production. However the particularity of our production mode is that we have one non eager laoded path (which I want to get rid of).

fxn:
I don't understand it. If all is equal, 2.1.0 should use less memory because `loaded_cpaths` is gone, and `to_unload` is empty. If `loaded_cpaths` had 1000 strings in 2.0.0, you have 1000 strings less in 2.1.0.

How strange...

casperisfine:
Yeah, super strange, I just double validated, made builds were good an all.

I've often seen strange results from `memory_profiler` it can sometimes misattribute some memory. However here I checked the total retained and it matches:

Zeitwerk 1.4.3: `4.81 MB  zeitwerk-1.4.3 / Total retained:  601.18 MB (4745405 objects)`
Zeitwerk 2.0.0: `2.74 MB  zeitwerk-2.0.0 / Total retained:  599.09 MB (4710814 objects)` 
Zeitwerk 2.1.0:  `3.39 MB  zeitwerk-2.1.0 / Total retained:  600.67 MB (4712388 objects)`

(please don't judge these numbers).

I got a couple pressing things, but I'll try to see if I can identify where this extra memory is coming from. 

That being said `1MB` isn't really a big concern, even a total 5MB should have been more than fine in production for us, so what I'd really like to figure out is what was causing memory to grow on each processed request (if it wasn't fixed by your recent patches). I'm suspecting `DescendantTracker`.

casperisfine:
So I've just put Zeitwerk to 5% of our production server, and I'm doing to heap analysis. I'll likely update later with more conclusions, but first a few observations:

  - Zeitwerk workers no longer grow their memory usage faster than regular workers. I think whatever was causing this was fixed either in Zeitwerk, or either in our app (we've made lots of work arounf that recently). So at this point there isn't anything really preventing us to use Zeitwerk in production.
  - Zeitwerk enabled unicorn master use ~84MB more memory than classic ones. I know where some of that is coming from, more on this later. 

Again, it's still very early so I can't say for sure where these 84MB are, however the head analysis show that we still have quite a lot of strings in the loaders:

```
harb> print 0x55726c87b6b0
    0x55726c87b6b0: "OBJECT"
              name: Zeitwerk::Loader
              size: 256
     references to: [
                      0x55726c87b688 (DATA: time)
                      0x5572693bf868 (STRING: "rails.main")
                      0x55726c87b7c8 (MODULE: Shopify::Application::ZeitwerkInflector)
                      0x55726c87b4a8 (HASH: size 461)
                      0x55726c87b480 (ARRAY: size 0)
                      0x55726c87b458 (OBJECT: Set)
                      0x55726c87b408 (OBJECT: Set)
                      0x55726c87b3b8 (HASH: size 74)
                      0x55726c87b390 (ARRAY: size 0)
                      0x55726c87b368 (HASH: size 0)
                      0x55726c87b340 (HASH: size 8)
                      0x55726c87b318 (HASH: size 2422)
                      0x55726c87b2a0 (OBJECT: Set)
                      0x55726c87b1d8 (DATA: mutex)
                      0x55726c87b1b0 (DATA: mutex)
```

The `2422` is `shadowed_files`: `Rails.autoloaders.main.shadowed_files.size => 2422`.

The `74` is `autoloads`: `Rails.autoloaders.main.autoloads.size => 74`

The `461` is `root_dirs` (I know...) `Rails.autoloaders.main.root_dirs.size => 461`

So our main issue seem to come from `shadowed_files`, this is likely because we still have a bunch of things loaded by the classic autoloader. I'm still working on reducing these.

I'll keep looking to make sure there isn't other places.

casperisfine:
Also, the only other loader is totally empty, which is great:

```
harb> print 0x55726c87b138
    0x55726c87b138: "OBJECT"
              name: Zeitwerk::Loader
              size: 232
     references to: [
                      0x55726c87b110 (DATA: time)
                      0x5572693bf638 (STRING: "rails.once")
                      0x55726c87b7c8 (MODULE: Shopify::Application::ZeitwerkInflector)
                      0x55726c87ae68 (HASH: size 0)
                      0x55726c87ae40 (ARRAY: size 0)
                      0x55726c87ae18 (OBJECT: Set)
                      0x55726c87adc8 (OBJECT: Set)
                      0x55726c87ad50 (HASH: size 0)
                      0x55726c87ad28 (ARRAY: size 0)
                      0x55726c87ad00 (HASH: size 0)
                      0x55726c87acb0 (HASH: size 0)
                      0x55726c87ac60 (HASH: size 0)
                      0x55726c87ac38 (OBJECT: Set)
                      0x55726c87abc0 (DATA: mutex)
                      0x55726c87ab98 (DATA: mutex)
                    ]
```

casperisfine:
Some analysis on one const name & path pair that is duplicated:

```
$ grep '"UnitSystem"' zeitwerk-master-heap.dump 
{"address":"0x557269d1a020", "type":"STRING", "class":"0x5572684f7c40", "bytesize":10, "capacity":127, "value":"UnitSystem", "encoding":"US-ASCII", "memsize":168, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x55726aba79f8", "type":"STRING", "class":"0x5572684f7c40", "embedded":true, "bytesize":10, "value":"UnitSystem", "encoding":"US-ASCII", "memsize":40, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x55726adf1790", "type":"CLASS", "class":"0x55726adf1768", "name":"UnitSystem", "references":["0x55726aded6b8", "0x55726ade9590", "0x55726ade96a8", "0x55726adec498", "0x55726adeb5c0", "0x55726adea0a8", "0x55726adecfd8", "0x55726ade9630", "0x55726ade95e0", "0x55726ade9608", "0x55726adeaa30", "0x55726adeea90", "0x55726adee1f8", "0x55726ade9658", "0x55726bc8ddd0", "0x55726c9f40f0", "0x55726adf8270", "0x55726adf1588", "0x55726adf8270", "0x55726adf13f8", "0x55726adf8270", "0x55726adf0908", "0x55726adf8270", "0x55726adf0840", "0x55726adf8270", "0x55726adf07f0", "0x55726adf8270", "0x55726adf0318", "0x55726adf8270", "0x55726adefc88", "0x55726adf8270", "0x55726adefc60", "0x55726adf8270", "0x55726adefc38", "0x55726adf8270", "0x55726adf0890", "0x55726adf8270", "0x55726adefbe8", "0x55726adf8270", "0x55726adf0e30", "0x55726adf8270", "0x55726adf16f0"], "memsize":1504, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x55726bc8ddd0", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "embedded":true, "fstring":true, "bytesize":10, "value":"UnitSystem", "encoding":"US-ASCII", "memsize":40, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
$ grep '"/app/app/models/unit_system.rb"' zeitwerk-master-heap.dump 
{"address":"0x55726aba8178", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "bytesize":30, "capacity":127, "value":"/app/app/models/unit_system.rb", "encoding":"US-ASCII", "memsize":168, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x55726adf2898", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "fstring":true, "bytesize":30, "value":"/app/app/models/unit_system.rb", "encoding":"US-ASCII", "memsize":71, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x5572ad068518", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "bytesize":30, "value":"/app/app/models/unit_system.rb", "encoding":"UTF-8", "memsize":71, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}


harb> rootpath 0x557269d1a020

root path to 0x557269d1a020:
                      ROOT (machine_context)
                      0x55726c877f60 (NONE: (null))
                      0x5572684fd730 (CLASS: Object)
                      0x55726adbe4f8 (MODULE: GraphApi)
                      0x557269ef4d00 (CLASS: GraphApi::UnitSystem)
                      0x557269d1a188 (OBJECT: GraphQL::EnumType)
                      0x557269d1a020 (STRING: "UnitSystem")

harb> rootpath 0x55726aba79f8

root path to 0x55726aba79f8:
                      ROOT (vm)
                      0x55726c88ff70 (DATA: tracepoint)
                      0x55726c88ff98 (DATA: proc)
                      0x55726c8942f0 (MODULE: Zeitwerk::ExplicitNamespace)
                      0x55726c894020 (HASH: size 2)
                      0x55726c87b6b0 (OBJECT: Zeitwerk::Loader)
                      0x55726c87b318 (HASH: size 2422)
                      0x55726aba79f8 (STRING: "UnitSystem")

harb> rootpath 0x55726bc8ddd0

root path to 0x55726bc8ddd0:
                      ROOT (machine_context)
                      0x55726c877f60 (NONE: (null))
                      0x5572684fd730 (CLASS: Object)
                      0x55726adf1790 (CLASS: UnitSystem)
                      0x55726bc8ddd0 (STRING: "UnitSystem")

harb> rootpath 0x55726aba8178

root path to 0x55726aba8178:
                      ROOT (vm)
                      0x55726c88ff70 (DATA: tracepoint)
                      0x55726c88ff98 (DATA: proc)
                      0x55726c8942f0 (MODULE: Zeitwerk::ExplicitNamespace)
                      0x55726c894020 (HASH: size 2)
                      0x55726c87b6b0 (OBJECT: Zeitwerk::Loader)
                      0x55726c87b318 (HASH: size 2422)
                      0x55726aba8178 (STRING: "/app/app/models/unit_system.rb")

harb> rootpath 0x55726adf2898

root path to 0x55726adf2898:
                      ROOT (vm)
                      0x5572684d4218 (ARRAY: size 17073)
                      0x5572ad920f48 (ARRAY: size 25609)
                      0x55726adf8270 (STRING: "(null)")
                      0x55726adf2898 (STRING: "/app/app/models/unit_system.rb")

harb> rootpath 0x5572ad068518

root path to 0x5572ad068518:
                      ROOT (vm)
                      0x5572684ccbf8 (DATA: VM/thread)
                      0x55726af92298 (DATA: fiber)
                      0x5572acc85b58 (ARRAY: size 10564)
                      0x5572acc84848 (STRING: "(null)")
                      0x5572ad068518 (STRING: "/app/app/models/unit_system.rb")
```

So they are both duplicated 3 times.

Class name:
  - Held in graphql metadata, not really related to Zeitwerk, always been there, and I have a PR en route to intern it.
  - Held in `Zeitwerk::Loader.shadowed_paths`, as we already figured
  - Held in the class itself, probably coming from [`ActiveModel::Naming`](https://github.com/rails/rails/blob/faf07d1468af06bb3f7f5dd0776d77dd252af3b6/activemodel/lib/active_model/naming.rb)

Might be worth deduplicating these.

Path:

  - Held in `Zeitwerk::Loader.shadowed_paths`, as we already figured
  - Held in `$LOADED_FEATURES`, expected (we're sure of it because `"fstring":true`)
  - Held in another global array, which very much look like `$LOADED_FEATURES`, but that one isn't interned. Very weird.

I'll try to identify what that last one is, but no clue so far.

Tangentially, there's also similar paths being held by `AS::Dependency` and Bootsnap, as expected:

```
$ grep '"/app/app/models/unit_system"' zeitwerk-master-heap.dump 
{"address":"0x55726adf82c0", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "bytesize":27, "value":"/app/app/models/unit_system", "encoding":"US-ASCII", "memsize":68, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x55726adf8400", "type":"STRING", "class":"0x5572684f7c40", "frozen":true, "bytesize":27, "capacity":30, "value":"/app/app/models/unit_system", "encoding":"US-ASCII", "memsize":71, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}

harb> rootpath 0x55726adf82c0

root path to 0x55726adf82c0:
                      ROOT (machine_context)
                      0x55726c877f60 (NONE: (null))
                      0x5572684fd730 (CLASS: Object)
                      0x55726ab8a948 (MODULE: ActiveSupport)
                      0x557269409c88 (MODULE: ActiveSupport::Dependencies)
                      0x5572694119b0 (OBJECT: Set)
                      0x557269411988 (HASH: size 2396)
                      0x55726adf82c0 (STRING: "/app/app/models/unit_system")

harb> rootpath 0x55726adf8400

root path to 0x55726adf8400:
                      ROOT (machine_context)
                      0x55726c877f60 (NONE: (null))
                      0x5572684fd730 (CLASS: Object)
                      0x55726b89ccf8 (MODULE: Bootsnap)
                      0x55726b894e18 (MODULE: Bootsnap::LoadPathCache)
                      0x55726b7b3a08 (OBJECT: Bootsnap::LoadPathCache::LoadedFeaturesIndex)
                      0x55726b7b39e0 (HASH: size 32639)
                      0x55726adf8400 (STRING: "/app/app/models/unit_system")


```


### Questions & Conclusions

  - How important is `shadowed_files`? Any way we could do without? I'll do my best to trim it down, but if we could simply not have it it would be awesome.
  - I'll submit a Rails PR to intern `ActiveModel::Naming` strings.
  - What are your thoughts on a simpler verison of this PR, to apply `String#-@` on constant names and paths?
  - With Zeitwerk registering `autoload` and calling `required` with absolute paths, I wonder if we could trim Bootsnap to only keep track of gems.

casperisfine:
> * How important is `shadowed_files`? Any way we could do without? I'll do my best to trim it down, but if we could simply not have it it would be awesome.

To answer my own question it's used in `eager_load`: https://github.com/fxn/zeitwerk/blob/b82a13c1526eb9184b1ab0796886efe9896b9a18/lib/zeitwerk/loader.rb#L354-L355

It prevent us from requiring constants that are already defined.

So I do think it could be dropped once `eager_load` finishes. @fxn is that right?

casperisfine:
> * Held in another global array, which very much look like `$LOADED_FEATURES`, but that one isn't interned. Very weird.

Ok, so it's actually held by two arrays, the weird global one (which I suspect to be some MRI ISeq registry), and also by `ActiveSupport::FileUpdateChecker`:

```
harb> rootpath 0x5572acc85ba8

root path to 0x5572acc85ba8:
                      ROOT (machine_context)
                      0x55726b9d3950 (OBJECT: Unicorn::HttpServer)
                      0x55726914c108 (OBJECT: PrereadInput)
                      0x55726c91dbe0 (OBJECT: Shopify::Application)
                      0x55726c91db18 (ARRAY: size 3)
                      0x5572acbae270 (OBJECT: ActiveSupport::FileUpdateChecker)
                      0x5572acc85ba8 (ARRAY: size 10564)
```

fxn:
Hey Jean, fantastic research! Just a heads up, travelling to Verona for RubyDay, will write back!

casperisfine:
No rush, I have a few idea who to improve this already.

Also: break a leg (for the conf)

casperisfine:
> and also by `ActiveSupport::FileUpdateChecker`:

Just figured I already had a PR for that one: https://github.com/rails/rails/pull/35701 -_-

fxn:
Quick followup, I am applying axe medicine to `shadowed_files`.

Its only purpose was to skip those on eager loading (but they are not autoloadable, so `do_not_eager_load` cannot be used for these ones).

I am clearing them after eager load, they are useless afterwards.

Also, I realize that the structure has a mixed debugging purpose, we do not really need the map. Instead, I am going to use a set and following that design guideline troubleshooting => logging I am going to log the shadowed file and reason if logging is enabled.

You can count on that one today.

casperisfine:
> You can count on that one today.

Awesome, but again no rush, Zeitwerk memory usage is no longer a blocking issue for us. If it takes a few days it's really fine.

fxn:
Version 2.1.1 is out with the optimized `shadowed_files`, I'll eventually continue the discussion!

casperisfine:
I just deployed 2.1.1, the unicorn masters are now `~8MB` lighter than with `2.1.0`. 🎉 

I'll keep looking for other reclaimable stuff.

fxn:
@casperisfine do you see potential in interning any strings now?

casperisfine:
@fxn I need to do another round of heap analysis now that we run Zeitwerk 2.1.2. I'll kickstart that right now.

casperisfine:
Looking at the same strings than before:

```
$ grep '"UnitSystem"' zeitwerk-2.1.2-master-dump 
{"address":"0x563efb379458", "type":"CLASS", "class":"0x563efb379430", "name":"UnitSystem", "references":["0x563efb370718", "0x563efb36e508", "0x563efb36e580", "0x563efb374570", "0x563efb373738", "0x563efb36e5d0", "0x563efb3720e0", "0x563efb375c68", "0x563efb3715f0", "0x563efb36fd40", "0x563efb36e4b8", "0x563efb36e4e0", "0x563efb372b58", "0x563efb36e5a8", "0x563efbd02a88", "0x563efca838b0", "0x563efb377d60", "0x563efb381068", "0x563efb3786e8", "0x563efb381068", "0x563efb381068", "0x563efb379278", "0x563efb381068", "0x563efb379070", "0x563efb381068", "0x563efb377d88", "0x563efb381068", "0x563efb377d38", "0x563efb381068", "0x563efb377d10", "0x563efb381068", "0x563efb3777c0", "0x563efb381068", "0x563efb376d48", "0x563efb381068", "0x563efb376d20", "0x563efb381068", "0x563efb376cf8", "0x563efb381068", "0x563efb376cd0", "0x563efb381068", "0x563efb379390"], "memsize":1504, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x563efbd02a88", "type":"STRING", "class":"0x563ef84cfc38", "frozen":true, "embedded":true, "fstring":true, "bytesize":10, "value":"UnitSystem", "encoding":"US-ASCII", "memsize":40, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x563f0a693c38", "type":"STRING", "class":"0x563ef84cfc38", "bytesize":10, "capacity":127, "value":"UnitSystem", "encoding":"US-ASCII", "memsize":168, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
$ grep '"/app/app/models/unit_system.rb"' zeitwerk-2.1.2-master-dump 
{"address":"0x563efb37a740", "type":"STRING", "class":"0x563ef84cfc38", "frozen":true, "fstring":true, "bytesize":30, "value":"/app/app/models/unit_system.rb", "encoding":"US-ASCII", "memsize":71, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
{"address":"0x563f42efa538", "type":"STRING", "class":"0x563ef84cfc38", "frozen":true, "bytesize":30, "value":"/app/app/models/unit_system.rb", "encoding":"UTF-8", "memsize":71, "flags":{"wb_protected":true, "old":true, "uncollectible":true, "marked":true}}
```

The class name string is only present twice (down from 4).

The first one is an interned string (`"fstring": true`), and is referenced from a bunch of places so that's all good:

```
harb> print 0x563efbd02a88
    0x563efbd02a88: "STRING"
             value: "UnitSystem"
              size: 80
   referenced from: [
                      0x563efb379458 (CLASS: UnitSystem)
                      0x563f125d6c70 (HASH: size 1507)
                      0x563f125d6db0 (HASH: size 1607)
                      0x563efbe6c770 (ARRAY: size 1024)
                    ]
harb> 
```

The second one is not interned, and is held by the GraphQL library, so not Zeitwerks fault:

```
harb> print 0x563f0a693c38
    0x563f0a693c38: "STRING"
             value: "UnitSystem"
              size: 208
   referenced from: [
                      0x563f0a693d50 (OBJECT: GraphQL::EnumType)
                    ]
```.

Now the paths:

The first one is an interned string (`"fstring": true`), and is held by `$LOADED_FEATURES` among many other things. 

As for the second one, it's frozen but not interned, and is held by that strange fiber/array that I've never been able to locate:

```
harb> rootpath 0x563f42efa538

root path to 0x563f42efa538:
                      ROOT (vm)
                      0x563ef84a4bf0 (DATA: VM/thread)
                      0x563efafca6e0 (DATA: fiber)
                      0x563f42cbc500 (ARRAY: size 10691)
                      0x563f42cab048 (STRING: "(null)")
                      0x563f42efa538 (STRING: "/app/app/models/unit_system.rb")
```

There's about 10k paths in there, I suspect it's some $LOADED_FEATURES cache data, I'd need to look into MRI source to figure that one out.

Either way to answer your initial question: No I don't see much opportunity for interning anymore. The class name held by Zeitwerk could be, but now that they are free after loading, interning them would only very marginally reduce memory usage in reloadable environments (likely not prod, so not that appealing) or during the boot process (really no huge deal as long as it's released at the end).

Thanks a lot for all that patching!

**Comment by casperisfine** [comment]
How and I forgot, here's what the main loader now look like:

```
harb> print 0x563ef9c72bd8
    0x563ef9c72bd8: "OBJECT"
              name: Zeitwerk::Loader
              size: 256
     references to: [
                      0x563ef9c72bb0 (DATA: time)
                      0x563ef9364c58 (STRING: "rails.main")
                      0x563ef9c72d18 (MODULE: Shopify::Application::ZeitwerkInflector)
                      0x563ef9c728e0 (HASH: size 461)
                      0x563ef9c72890 (ARRAY: size 0)
                      0x563ef9c72868 (OBJECT: Set)
                      0x563ef9c72818 (OBJECT: Set)
                      0x563ef9c727c8 (HASH: size 70)
                      0x563ef9c727a0 (ARRAY: size 0)
                      0x563ef9c72778 (HASH: size 0)
                      0x563ef9c72750 (HASH: size 9)
                      0x563ef9c72728 (OBJECT: Set)
                      0x563ef9c726d8 (OBJECT: Set)
                      0x563ef9c72688 (DATA: mutex)
                      0x563ef9c72660 (DATA: mutex)
                    ]
```

`HASH: size 461` is our load path (I know...)

`0x563ef9c727c8 (HASH: size 70)` is the last few non eager loaded constants, and a few other stuff I need to cleanup. So that's on us, not Zeitwerk

fxn:
That is great, awesome collaboration in this thread Jean!

PR #35, opened 2019-04-10
Fix "warning: shadowing outer local variable - dir"

y-yagi:
This warning has been removed in Ruby 2.6. But now this gem supports lower version, so I think it should be fixed.


fxn:
Thanks!

PR #36, opened 2019-04-10
Clear Loader#shadowed_files after eager_load

casperisfine:
As discussed in https://github.com/fxn/zeitwerk/pull/34#issuecomment-481606665, I think `shadowed_files` can be cleared when `eager_load` is called.

As far as I can tell `eager_load` is the only place where we query that data structure, and it can only be effectively called once.

Note that this change breaks lots of tests that use it to make assertions, I can likely fix them but I'd rather do it after I got some 👍 otherwise it's just wasted effort.

cc @fxn


casperisfine:
❤️

PR #40, opened 2019-04-19
Use caller_locations to get caller filepath

hkdnet:
If `Zeitwerk::Loader.for_gem` is called from a file which contains `:` in its path, `called_from` will be misunderstood.
We can get correct filenames via `caller_locations`.

```ruby
# /tmp/tm:p.rb
require 'zeitwerk'

puts loader.tag # => tm
```

I know this rarely happens in actual because bundle path and gem name cannot contain `:`. But my code seems somehow more readable.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -413,7 +413,7 @@ module Zeitwerk
       #
       # @return [Zeitwerk::Loader]
       def for_gem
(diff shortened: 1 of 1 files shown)


fxn:
Guess that would be important in Windows, where the patch may be (I guess) `C:/...`.

Thanks!

hkdnet:
Oh, I missed Windows!
Thanks for your quick reply :)

PR #41, opened 2019-04-19
Fix document on Inflector usage

aeroastro:
Since `Zeitwerk::Inflector#camelize` is an instance method, I have fixed the document to follow the code.


fxn:
Good catch! I believe I refactored it sometime during development and docs remained outdated.

Thank you!

PR #44, opened 2019-04-26
Avoid building the entire stack to get the caller path

casperisfine:
This is a fairly minor performance patch.

`caller` and `caller_location` will allocate one object per stack frame (a `String` and a `Thread::Backtrace::Location` respectively).

But when you know you only want to look at a subset of the call stack, you can pass `start` and `length` arguments, e.g. `caller(0, 1)` to only get the last frame.

No huge deal here as `for_gem` is not really a hotpsot, nor likely to be called from very deep, but the change is so simple that I figured I'd submit it.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -414,7 +414,7 @@ module Zeitwerk
       #
       # @return [Zeitwerk::Loader]
       def for_gem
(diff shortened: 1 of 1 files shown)


fxn:
Oh, absolutely, I prefer the code to be precise like that.

PR #46, opened 2019-04-30
Handle namespaces being in both eager loaded and non-eager loaded paths

casperisfine:
I found a minor issue with `2.1.5` eager loading. The test case should explain it more precisely but in short:

  - A same namespace can be set to be autovivified from multiple sub directories
  - Only the first one found is registered in `autoloads`, other ones are stored in `lazy_subdirs`.
  - If the first path to be found is in the `eager_load` exclusion list, `eager_load` will ignore all the other directories that should autovivify the namespace. Hence sub constants won't be eagerloaded even though they should.

cc @rafaelfrance @Edouard-Chin @fxn

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -481,6 +481,7 @@ module Zeitwerk
         # no matter if it is for a file or a directory. Just remember the
         # subdirectory has to be visited if the namespace is used.
         (lazy_subdirs[cpath] ||= []) << subdir
+        autoloads[subdir] = [parent, cname]
       elsif !cdef?(parent, cname)
         # First time we find this namespace, set an autoload for it.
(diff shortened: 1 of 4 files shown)


mensfeld:
This is a problem for me as well.

fxn:
ACK!

Let me think about it. The `autoloads` hash contains data about actual `autoload` calls, if we throw all subdirectories there it changes the meaning and uses more memory. Maybe that's what we need to do, but prefer to give it a thought.

By the way, I just pushed today

```
loader.log!
```

to print to `$stdout`, I have written `= method(:puts)` enough (and probably you too :).

fxn:
I restored `require` based eager loading. We can do that know because the expected inflection is checked.

Have published 2.1.6 including this fix.

PR #58, opened 2019-06-27
Add a bit more info about eager loading

woarewe:
Hello.
I started using **zeitwerk** in project which has different custom CLI's.
Corresponded code base is in dir ```$PROJECT_ROOT/lib/cli```.
I played with code loading and in first version of boot file uses just ```eager_load``` without any inflectors
I started application and to my surprise it works correctly. All constants and namespaces were loaded right.
Then I removed ```eager_load``` method calling, restart my app, and, as expected, I got error:

```shell
Traceback (most recent call last):
bin/template:6:in `<main>': uninitialized constant CLI (NameError)
Did you mean?  Cli
```

I have fully understanding about runtime loading, why I need inflectors, but I didn't understand why I don't need them when I using eager loading? Using **pry-byebug** I started studying zeitwerk realisation and found that it uses require for eager loading.

I think, like a gem user I shouldn't do so much actions to learn how it works, much easier if it will be in documentation or README.

If you are not agree with me, just close this PR.
Thanks for attention, have a nice day!

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ loader.eager_load # won't eager load the database adapters
 ```
 
 Eager loading is synchronized and idempotent.
+Eager loading doesn't need inflectors due to using ```require``` for files loading.
 
 If you want to eager load yourself and all dependencies using Zeitwerk, you can broadcast the `eager_load` call to all instances:
 


fxn:
Sounds like a bug to me. If the name of the file or directory does not match inflection, [you should get an error](https://github.com/fxn/zeitwerk/blob/ea186a51cf6989d097cf3ce5d740a697c482ac35/lib/zeitwerk/loader/callbacks.rb#L15).

Could you please get a minimal way to reproducem enable logging with

```ruby
loader.log!
```

before setup, and share the repo and the traces?

woarewe:
Maybe I described situation a bit intricately, my message is about using eager loading without inflectors, it works, but why?) Without investigation it is not clear.
My repo link
https://github.com/woarewe/clang_journey/blob/others/zeitwerk-situation-reproduction/config/boot.rb#L15
I created another branch to reproduce described context.

fxn:
Yes, yes, I understood.

My answer is that it looks like a bug. If you eager load successfully, you should be able to autoload the same project successfully. If you cannot, that is an issue that needs code fixing rather than docs.

(That said without having investigated the actual problem yet.)

woarewe:
Do you mean, zeitwerk should also check inflections when eager loading before files requiring and raise exception if something don't match?

fxn:
Exactly. It already does indeed, but your project might have hit an overlooked case.

The way it works is that `Kernel.require` is decorated by the gem. If the file being required is one managed by Zeiwerk, [an internal callback is invoked]. That is the one I linked above, and among other things it verifies the expected constant is defined.

fxn:
Just published 2.1.7 including this fix.

PR #62, opened 2019-07-06
Use `Class#name` to compute constant name

tricknotes:
The user defined class/module's `name` may be overridden.

### DIFF
--- a/lib/zeitwerk/explicit_namespace.rb
+++ b/lib/zeitwerk/explicit_namespace.rb
@@ -63,7 +63,7 @@ module Zeitwerk
 
       # Note that it makes sense to compute the hash code unconditionally,
       # because the trace point is disabled if cpaths is empty.
-      if loader = cpaths.delete(event.self.name)
+      if loader = cpaths.delete(Class.instance_method(:name).bind(event.self).call)
         loader.on_namespace_loaded(event.self)
         disable_tracer_if_unneeded
       end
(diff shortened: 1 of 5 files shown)


fxn:
I understand the theoretical reason for this.

But the fact that users can do formally something weird, is not enough for me to consider supporting that. To put an extreme example of what I mean, users can technically remove the `String` constant from `Object`, the API allows that, yes, but programs just won't work.

So, to consider this pull request, I'd need a compelling use case.

casperisfine:
I'd add that this way of accessing the module name if certainly much less efficient, so if you ever consider merging this, please let me know and I'll run some benchmark.

tricknotes:
Thanks for your response, @fxn
Also, I'm sorry for my too less explanation.

I found this problem from real use case.
My application uses [settingslogic](https://github.com/binarylogic/settingslogic) to manage settings.
This gem overrides [`Class#name`](
https://github.com/binarylogic/settingslogic/blob/v2.0.8/lib/settingslogic.rb#L9) and it returns setting variables.
I think this is a bit tricky usage but it makes sense inside this gem's API.

Yes, I know this gem is not maintained for now.
However, I think that the usage like this is not so extremely unusual.

fxn:
It's OK.

I have also verified that `inspect` and `to_s` ignore `name`, in MRI they are aliased and seem to go to directly a table of constant paths cached on the first constant assignment.

Working on this.

fxn:
Thanks! ❤️

fxn:
Hey @tricknotes, 2.1.9 was just published including this patch.

tricknotes:
Thanks! ✨

PR #64, opened 2019-07-11
Stabilize test suite

kvokka:
Hello,

A few tests produced red status (Ruby 2.6.3), but only occasionally, once in 5 or 10 runs.

It was produced by leaking tests- some of them did not clean up well after themselves. I found the problem and fixed it.


fxn:
Fixed by https://github.com/fxn/zeitwerk/commit/98b556a7486ba5e3d9d5e0cd55dd4350590d4509. Only that test was leaking.

Issue #65, opened 2019-07-11
Loader error on ignored paths


kvokka:
Steps to reproduce:

```
# my_gem
loader = Zeitwerk::Loader.for_gem
loader.ignore 'lib/foo/bar'
loader.setup
```

Later, in another file:
```
# lib/foo/bar/zoo/my_file
loader = Zeitwerk::Loader.new
loader.push_dir "#{__dir__}/active_record" # <= raised here
loader.setup
```

It looks like the problem is in `Loade::raise_if_conflicting_directory(dir)` method. Don't you mind if I'll take it for PR?

fxn:
I don't fully undersand. Are you instantiating a loader inside an ignored path of the same project?

kvokka:
Thank you for a such a quick response.

Sorry, it is looking like the initial description was clumsy.

Yes. I'm Implementing the adapter pattern (in the gem, but it does not the case). I followed this https://github.com/fxn/zeitwerk#use-case-the-adapter-pattern part of the documentation, and it was ended with `require` which loads the adapter. But the adapter may load some assets specific for this adapter. And I want to load this assets with another instance of zeitwerk (this path was ignored in the very beginning by main loader).

At this moment I realized, that it is really edge use case.

I'll try to handle it. Don't you mind if I also add this case in the README?

fxn:
It would help me understand the use case better if you could share some real files. At least a bunch of real file names (contents not needed). I get the idea of what you're doing, but that would be even better if possible.

kvokka:
No problem

```ruby
# lib/active_validation.rb
loader = Zeitwerk::Loader.for_gem
loader.ignore 'lib/active_validation/frameworks'
loader.setup
```

```ruby
# lib/active_validation/orm_adapters/active_record.rb
# The adapter file
...

# If the adapter is loaded, it load the assets
require 'active_validation/frameworks/active_record'
```

```ruby
# lib/active_validation/frameworks/active_record.rb
loader = Zeitwerk::Loader.new
loader.push_dir "#{__dir__}/active_record/models"
loader.setup
```

kvokka:
I made my tests green with local patch of `zeitwerk` gem, but did not write the tests for `zeitwerk` yet and the additions for the docs. Will try to deal with it today later and give you a PR

kvokka:
You closed my PR, so i have to write here.

[...]

PR #67, opened 2019-07-11
Fix the use case with adapter pattern

kvokka:
Fix for #65 issue.

For achieving stable green tests requires #64 (but this PR does not depends on it)

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,29 @@ The chosen adapter, then, has to be loaded by hand somehow:
 require "my_gem/db_adapters/#{config[:db_adapter]}"
 ```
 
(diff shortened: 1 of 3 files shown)


kvokka:
The CI as expected is red, but this example shows well, why you need #64, which fix it

fxn:
I believe the right way to address this is to improve the detection of conflicting root directories.

The purpose of that check is to prevent users from accidentally having two loaders managing the same files. They would step on each other.

But the current check is not taking into account ignored paths. Anything below an ignored path is in particular _not_ managed by that loader, so there is no conflict if a different loader wants to.

I'll close here, and will work on a fix for #65.

PR #68, opened 2019-07-14
Fix the usecase with adapter pattern

kvokka:
#67 

But with fixed readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -461,6 +461,28 @@ The chosen adapter, then, has to be loaded by hand somehow:
 require "my_gem/db_adapters/#{config[:db_adapter]}"
 ```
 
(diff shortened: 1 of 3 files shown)


fxn:
I have pushed https://github.com/fxn/zeitwerk/commit/15f53194cbb292027c0338b8e30c7c8b1e3e4f5a, implementing this.

Ignored paths were computed on setup, because they were not needed earlier. But now we need to expand patterns on the spot to behave as with the root directories in conflict detection, since root directories are checked regardless of whether the loader has called `setup`.

Additionally, conflict detection started to feel a bit too much, and have encapsulated the condition in the new private predicate `manages?`.

Tests are simplified too.

You get of course credit for your contribution to this fix.

fxn:
I just published 2.1.9 including that patch.

kvokka:
Thank you! This patch helped and looks cleaner.

PR #69, opened 2019-07-15
Refactor ExplictlNamespace tracepoint to use a method rather than a block

casperisfine:
To be honest the main justification for this change is "cosmetic". The TracePoint is one of the main Zeitwerk hotspots, and as such often appear in profiles under the `block in <module:ExplicitNamespace>` name.

I'd prefer if it had a more descriptive name.

There's also probably some extremely marginal performance gain by getting rid of a closure in favor of a method, but it's likely so small that I don't think it's worth benchmarking.

### DIFF
--- a/lib/zeitwerk/explicit_namespace.rb
+++ b/lib/zeitwerk/explicit_namespace.rb
@@ -54,22 +54,24 @@ module Zeitwerk
           tracer.disable if cpaths.empty?
         end
       end
+
+      def tracepoint_class_callback(event)
+        # If the class is a singleton class, we won't do anything with it so we
+        # can bail out immediately. This is several orders of magnitude faster
+        # than accessing its name.
+        return if event.self.singleton_class?
+
+        # Note that it makes sense to compute the hash code unconditionally,
+        # because the trace point is disabled if cpaths is empty.
+        if loader = cpaths.delete(real_mod_name(event.self))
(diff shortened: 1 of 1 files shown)


fxn:
👍

The TP is called when classes/modules are created, and also when they are reopened. The method name looks almost fine, but not 100% for my taste. Could we imagine another name that is useful for your purpose?

casperisfine:
For sure, that's just the first one that came to me.

One possibility would simply to reuse `on_namespace_loaded`, but I suppose `namespace` means "zeitwerk tracked namespace".

But if not, then it could be `on_namespace_defined`, or `on_module_opened` / `on_namespace_opened`.

I'll keep thinking about other names.

fxn:
Would it make sense to call it `zeitwerk_tp_callback` or something like that? Just to throw an idea.

casperisfine:
> `zeitwerk_tp_callback` or something like that?

Yeah it totally could. As in not necessarily that exact name, but a very technical term: e.g.`tracepoint_class_callback`.

I don't think the `zeitwerk_` prefix is necessary.

fxn:
👍

casperisfine:
Updated with `tracepoint_class_callback`.

fxn:
Hey Jean, 2.1.9 is out including this.

Issue #73, opened 2019-08-19
const_defined?': wrong constant name Font-awesome (NameError)


medicharlachiranjeevi:
Hi we are using rails version 6.0.0 and when i run using rails version 6.0.0.rc1 its working fine
Error log 
`
home/system/.rbenv/versions/2.6.1/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:705:in `const_defined?': wrong constant name Font-awesome (NameError)
`

fxn:
Can you please provide a minimal app to reproduce?

fxn:
Can you please share the log of the loader? Just throw

```ruby
Rails.autoloaders.log!
```

in `config/application.rb` and trigger the error.

Where is the directory `font-awesome` located?

medicharlachiranjeevi:
thanks for reply in app/pdf/shared/stylesheets

fxn:
OK, that directory should not be in the autoload paths.

Do you have Ruby code to be autoloaded below `app/pdf` at all?

medicharlachiranjeevi:
Thanks it worked

fxn:
Version 2.1.10 is out an includes that change.

Issue #72, opened 2019-08-19
Zeitwerk trying to autoload YAML files since 2.1.9


thelucid:
Since upgrading to Rails 6 from 6.rc2, I'm getting an error whereby Zeitwerk is truing to autoload a YAML file placed in `my-rails-app/some-dir/some-subdir/1-my-file.yml.

Framework trace is:

```
NameError: wrong constant name 1-my-file
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:705:in `const_defined?'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:705:in `cdef?'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:610:in `strict_autoload_path'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:587:in `autoload_for?'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:501:in `autoload_subdir'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:490:in `block in set_autoloads_in_dir'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:662:in `block in ls'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:659:in `foreach'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:659:in `ls'
~/.rbenv/versions/2.6.3/lib/ruby/gems/2.6.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:476:in `set_autoloads_in_dir'
[...]

PR #75, opened 2019-08-23
Better error for invalid constant names

casperisfine:
Ref: https://github.com/fxn/zeitwerk/issues/72 https://github.com/fxn/zeitwerk/issues/73

People seem to be confused by this error. However it's a totally legitimate error case, and after playing with it a bit I don't think automatically silencing those is a good idea, because if you customize the inflector and include a bug in it, you might have a hard time figuring it out.

However I think what we can do is to improve the error message, so that users better understand what happened, and can fix the issue themselves without opening an issue on GitHub.

NB: doc/writing isn't exactly my *forte*, so alternative error messages are welcome.

Also I chose to subclass `::NameError`, so that it's clearer that it's about Zeitwerk, and also to allow Rails (arguably the main way people we use Zeitwerk) to provide better contextual help.

@fxn WDYT?

### DIFF
--- a/lib/zeitwerk/error.rb
+++ b/lib/zeitwerk/error.rb
@@ -4,4 +4,7 @@ module Zeitwerk
 
   class ReloadingDisabledError < Error
   end
+
+  class NameError < ::NameError
+  end
 end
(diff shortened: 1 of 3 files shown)


fxn:
Slowly cacthing up with Life(tm).

Thanks a lot for the patch Jean. It was good, and I think subclassing `NameError` is a great idea. I added the explicit `begin` and tweaked just a little bit the message.

Issue #78, opened 2019-08-31
Error running an example from synopsis


bruno-:
Hi,

I'm playing with this gem by running an example from synopsis:

```ruby
# lib/my_gem.rb
require "zeitwerk"
loader = Zeitwerk::Loader.for_gem
loader.setup

module MyGem
end

loader.eager_load
```

By running `ruby lib/my_gem.rb` I'm getting the below error.

- `ruby --version`: `ruby 2.5.5p157 (2019-03-15 revision 67260) [x86_64-darwin17]`
- `gem list zeitwerk`: `zeitwerk (2.1.9)`

Am I doing something wrong? Thanks!

```
Traceback (most recent call last):                 
        14: from lib/my_gem.rb:5:in `<main>'       
        13: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.9/lib/zeitwerk/kernel.rb:16:in `require'
        12: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        11: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require'
        10: from /Users/bruno/data/edu/project/lib/my_gem.rb:2:in `<top (required)>'                  
         9: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.9/lib/zeitwerk/loader.rb:442:in `for_gem'
         8: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.9/lib/zeitwerk/registry.rb:85:in `loader_for_gem'
         7: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.9/lib/zeitwerk/registry.rb:85:in `tap'
         6: from /Users/bruno/.rbenv/versions/2.5.5/lib/ruby/gems/2.5.0/gems/zeitwerk-2.1.9/lib/zeitwerk/registry.rb:88:in `block in loader_for_gem'
[...]

PR #79, opened 2019-09-01
Expand the caller path in Loader.for_gem

byroot:
Fix: https://github.com/fxn/zeitwerk/issues/78

I've put some debug statement, and in short what was happening is:

  - When running the code as `ruby my_gem.rb`, the caller path in the stack is relative ( `caller_locations(0, 1).first.path # => "my_gem.rb"`)
  - Because of this the loader would be registered in `Zeitwerk::Registry.loaders_managing_gems["my_gem.rb"]`.
  - However the autoloads and such would be registered with fully expanded path.
  - So when we reach `module MyGem` it triggers an autoload and load the same file again, causing all sorts of shenanigans.

IMO running a gem like this doesn't make much sense, however since the fix is very simple, and that there might be cases we didn't think about that might cause the caller path to be relative, I think it's a good idea to simply handle it properly.

@fxn @bruno-

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -438,7 +438,7 @@ module Zeitwerk
       #
       # @return [Zeitwerk::Loader]
       def for_gem
-        called_from = caller_locations(1, 1).first.path
+        called_from = File.expand_path(caller_locations(1, 1).first.path)
         Registry.loader_for_gem(called_from)
       end
 
--- a/test/lib/zeitwerk/test_for_gem.rb
+++ b/test/lib/zeitwerk/test_for_gem.rb
@@ -105,4 +105,31 @@ class TestForGem < LoaderTest
       end
     end
   end
+
+  test "is able to handle a caller defined in a relative path" do
+    on_teardown do
+      remove_const :MyGem
+      delete_loaded_feature "my_gem.rb"
+    end
+
+    files = [
+      ["my_gem.rb", <<-EOS],
+        $for_gem_test_loader = Zeitwerk::Loader.for_gem
+        $for_gem_test_loader.setup
+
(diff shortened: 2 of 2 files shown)


bruno-:
I applied the patch from this PR. I'm running it with this code (very similar to example from #78):

```ruby
# lib/my_gem.rb
require "zeitwerk"
loader.setup

module MyGem
  puts "Foo"
end

loader.eager_load
```

I'm running it with `ruby lib/my_gem.rb`. I'm not getting an error anymore. The output I get is below which is still a wee bit surprising as it indicates `lib/my_gem.rb` was loaded 2 times. There's a good chance I don't grok how this work and how it should be used.

```
Foo
Foo
```

byroot:
Yeah, I made this not break, but there is still double loading. As said on the issue you filed, `lib/my_gem.rb` being the entry point make very little sense. I'll see if I can find a way to handle it, but really that's a huge corner case, and I can't think of any legitimate reason to do it this way.

byroot:
Ok, I added some more code to prevent the double loading. But really this is getting ugly.

bruno-:
> I can't think of any legitimate reason to do it this way.

Roger that. I think it's my misconception about how things should work. I'm ok with handling this how the maintainers see fit...

byroot:
> handling this how the maintainers see fit...

Well, the assumption is that `my_gem.rb` is required, not the entry point of your program.

fxn:
As @byroot said, gem entry points are expected to be required, and loading it with Ruby directly is not supported.

The idiomatic way to do that is either

```
ruby -Ilib -rmy_gem ...
```

or, more commonly, define an executable entry point in the `bin` folder that requires the main file. Such an executable may ship with the gem or be used only for development.

PR #81, opened 2019-09-10
Route back to Trusty image

cotsog:


fxn:
For the archives.

Recent builds have been failing on `ruby-head` because the VM setup stalls at

```
importing gemset /home/travis/.rvm/gemsets/global.gems
```

and a timeout aborts the build.

Travis CI is migrating boxes from Trusty to Xenial, and they want to try whether that is related by fixing the distribution.

Let's see.

PR #88, opened 2019-10-05
Add Zeitwerk::SimpleInflector

timcraft:
Thanks for zeitwerk, much easier than managing requires manually!

I've integrated zeitwerk across a number of different projects and the only pain point I keep hitting is having to write custom inflectors. Whilst it's very flexible, having to write a custom 10+ line class just to handle a few acronyms and abbreviations is a bit onerous.

Given how common acronyms are (the standard library has CGI, CSV, DBM, DRB, ERB, GDBM, IRB, JSON, RSS, SDBM, URI, YAML), I think there is value in including an inflector implementation that can eliminate the need to write a custom inflector for these simple cases.

As a potential solution this PR adds a Zeitwerk::SimpleInflector class which takes a hash of inflections for special cases. As well as basic acronyms this can also handle basenames ending with capitalized words (e.g. global_id), and irregularities like mysql/dynamodb.

Before:

```ruby
class MyInflector < Zeitwerk::Inflector
  def camelize(basename, _abspath)
    case basename
    when "api", "http"
      basename.upcase
    when "global_id"
      "GlobalID"
    when "mysql"
      "MySQL"
    else
      super
    end
  end
end

loader.inflector = MyInflector.new
```

After:

```ruby
loader.inflector = Zeitwerk::SimpleInflector.new({
  'api' => 'API',
  'http' => 'HTTP',
  'global_id' => 'GlobalID',
  'mysql' => 'MySQL'
})
```

Or alternatively:

```ruby
loader.inflector = Zeitwerk::SimpleInflector.new.tap do |inflector|
  inflector.inflect 'api', 'API'
  inflector.inflect 'http', 'HTTP'
  inflector.inflect 'global_id', 'GlobalID'
  inflector.inflect 'mysql', 'MySQL'
end
```

### DIFF
--- a/lib/zeitwerk.rb
+++ b/lib/zeitwerk.rb
@@ -7,6 +7,7 @@ module Zeitwerk
   require_relative "zeitwerk/explicit_namespace"
   require_relative "zeitwerk/inflector"
   require_relative "zeitwerk/gem_inflector"
+  require_relative "zeitwerk/simple_inflector"
   require_relative "zeitwerk/kernel"
   require_relative "zeitwerk/error"
 end
--- a/lib/zeitwerk/simple_inflector.rb
+++ b/lib/zeitwerk/simple_inflector.rb
@@ -1,0 +1,22 @@
+module Zeitwerk
+  class SimpleInflector < GemInflector # :nodoc:
+    # @param inflections [Hash, nil]
+    def initialize(inflections = nil)
+      @inflections = inflections || {}
+    end
+
+    # @param basename [String]
+    # @param class_name [String]
+    # @return [String]
+    def inflect(basename, class_name)
+      @inflections[basename] = class_name
+    end
+
+    # @param basename [String]
+    # @param abspath [String]
+    # @return [String]
+    def camelize(basename, _abspath)
+      @inflections[basename] || super
+    end
+  end
+end
--- a/test/lib/test_simple_inflector.rb
+++ b/test/lib/test_simple_inflector.rb
@@ -1,0 +1,32 @@
+require "test_helper"
+
+class TestSimpleInflector < Minitest::Test
+  def inflections
+    {
+      "api" => "API",
+      "global_id" => "GlobalID",
+      "mysql" => "MySQL",
+    }
+  end
+
+  test "camelizes using the inflections hash" do
+    inflector = Zeitwerk::SimpleInflector.new(inflections)
+
+    assert_equal "API", inflector.camelize("api", nil)
+    assert_equal "GlobalID", inflector.camelize("global_id", nil)
+    assert_equal "MySQL", inflector.camelize("mysql", nil)
+  end
+
+  test "camelizes basenames not in the inflections hash normally" do
+    inflector = Zeitwerk::SimpleInflector.new(inflections)
+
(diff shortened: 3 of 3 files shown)


fxn:
I like this interface. Experience shows you normally write a hash table/`case`/`if` expression that is just a map of strings. It is also very simple, backwards and forwards compatible.

Let me think about the exact details, but 👍.

fxn:
I believe we could add this to the current basic inflector. Usage would be:

```ruby
loader = Zeitwerk::Loader.new
loader.inflector.inflect(
  html_parser:    "HTMLParser",
  csv_controller: "CSVController",
  mysql_adapter:  "MySQLAdapter"
)
```

In the interface I am imagining `initialize` is left untouched, and `inflect` would receive a hash. I guess `inflect` could be called multiple times and would merge into an existing hash from the second call on. Off the top of my head:

```ruby
def inflect(inflections)
  (@__inflections ||= {}).merge!(inflections)
end
```

The implementation of `camelize` would not compute the hash code of `basename`  if there are no custom inflections (off the top of my head):

```ruby
def camelize(basename, _absname)
  (@__inflections && @__inflections[basename]) || basename.split('_').map!(&:capitalize).join
end
```

Regarding compatibility:

* Existing code using this inflector is compatible because the existing code does not invoke `inflect`.
* Subclasses of this inflector could define `inflect`, but they would override `camelize` anyway. The purpose of the double underscore is to avoid clashing with an existing `@inflections` and delegation to `super`.
* I believe we can do this because the change is simple, just adding a cache, no regexps or fancy logic that could need to be tweaked in future versions. If you want something else, then you implement your own inflector.

I prefer to have this in the base inflector because usage is saying this interface is convenient, and Zeitwerk is in its early days, so better put it right there thinking long term.

What do you think?

fxn:
Ah! We have to stringfy the keys of the inflections. Well, only outlining the idea, details filled when sitting down in the editor :).

timcraft:
> Experience shows you normally write a hash table/case/if expression that is just a map of strings. It is also very simple, backwards and forwards compatible.

Yes, so far I've just been using case expressions on the basename like in the README. Simple, but a bit verbose. I haven't yet needed to use abspath to disambiguate anything, which is why I think the 80/20 can be handled more easily with a hash mapping (or some equivalent code).

The other thing I considered was something even simpler for acronyms:

```ruby
loader.inflector = Zeitwerk::AcronymInflector.new(['api', 'http'])
```

But that wouldn't handle those other cases, which are common enough.

> I prefer to have this in the base inflector because usage is saying this interface is convenient, and Zeitwerk is in its early days, so better put it right there thinking long term.

Great, it would be even more convenient to have it in the base inflector as that would eliminate the additional line of code to set the custom inflector 👍

Symbol keys are nice and readable, but would require stringification as you say. I'm not sure what's best there (more code to maintain), could always add support for symbol keys at a later date.

fxn:
Let’s go with strings only 👍🏼.

timcraft:
> Let’s go with strings only 👍🏼.

Ok I've implemented that in #89, only change compared to your outline is adding a private reader method to avoid lots of `instance variable @__inflections not initialized` warnings.

fxn:
Closing in favor of #89.

PR #89, opened 2019-10-06
Add Zeitwerk::Inflector#inflect method

timcraft:
Adds a Zeitwerk::Inflector#inflect method which can be used to explicitly define custom inflections that take priority over the default Zeitwerk::Inflector#camelize behaviour, per https://github.com/fxn/zeitwerk/pull/88#issuecomment-538712084

Using a double underscored ivar to avoid clashes with existing `@inflections` and using a private reader method to prevent `instance variable @__inflections not initialized` warnings.

### DIFF
--- a/lib/zeitwerk/inflector.rb
+++ b/lib/zeitwerk/inflector.rb
@@ -13,7 +13,19 @@ module Zeitwerk
     # @param _abspath [String]
     # @return [String]
     def camelize(basename, _abspath)
-      basename.split('_').map!(&:capitalize).join
+      __inflections[basename] || basename.split('_').map!(&:capitalize).join
+    end
+
+    # @param inflections [Hash]
+    # @return [Hash]
+    def inflect(inflections)
+      __inflections.merge!(inflections)
+    end
+
+    private
+
+    def __inflections
+      @__inflections ||= {}
     end
   end
 end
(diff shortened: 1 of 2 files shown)


fxn:
❤️

fxn:
Hey! Just a heads to say I merged the patch because I believe this is needed. But I’ll think about the definitive version a bit.

On one hand, I am not fully convinced about the method name. We say `inflect`, but pass only a few rules, the rest are implicit. Doesn’t fully click to me.

That hash is configuring... fixed mappings that take precedence, overrides, special cases, ..., there is a more concise name in there.

On the other hand, I dislike the implementation. We’ve done it that way to try to be as backwards compat as possible, but it is not something I would write. So, even if the likelihood of a clash is really small, I am pondering to write regular code that initializes an ivar in the constructor, and publish 3.0.0. It is the right time to do it, and the API deserves it.

I need to think about it.

fxn:
After rubber ducking this quite bit, I could not find anything better than `inflect`, and have addressed the exact meaning by using words like "override" in the documentation.

The ivar is called `@overrides`, and I have left it lazy initialized as it was originally because not even the gem inflector called `super`. So, a compromise, we not have leading underscores, but the constructor is left untouched.

It is out as version 2.2.0.

Thanks for the patch again!

timcraft:
> After rubber ducking this quite a bit, I could not find anything better than `inflect`, and have addressed the exact meaning by using words like "override" in the documentation.

Yeah the naming is difficult as usual. It's not 100% clear but I couldn't think of anything better than `inflect` or `inflections`, `exceptions` would obviously be too confusing, and ActiveSupport's Inflector API is too much for what's needed here.

This will definitely make it smoother to integrate, thanks!

PR #93, opened 2019-10-21
Enable two optimization for newer rubies

casperisfine:
Both were added in MRI 2.7.

The first one is `Module#autoload?` now taking a second argument
to restrict the search to the module itself.
https://bugs.ruby-lang.org/issues/15777

The second is the new `UnboundMethod#bind_call` that allows
calling an unbound method without first allocating a `Method`
instance.
https://bugs.ruby-lang.org/issues/15955

@fxn 

cc @rafaelfranca @Edouard-Chin @etiennebarrie


fxn:
Fantastic! Thanks Jean!

PR #94, opened 2019-10-26
support auto-loading from specific module other than Object

xiaohui-zhangxh:
Hi, 

As Rails style, we usually place namespaced files like this:

```ruby
# app/models/user.rb
class User < ActiveRecord::Base
end

# app/forms/users/sign_up_form.rb
class Users::SignupForm
  include ActiveModel::Model
  attr_accessor :login, :password
  validates :login, :password, presence: true
end

# app/services/users/sign_up.rb
class Users::SignUp
  def call(params)
    form = Users::SignupForm.new(params)
    if form.valid?
      # ...
    else 
      # ...
    end
  end
end
```

Now, I'm going to refactor user-related business logics to individual directory, like this:

```ruby
# domains/users/models/user.rb
class User < ActiveRecord::Base
end

# domains/users/forms/sign_up_form.rb
class Users::SignupForm
  include ActiveModel::Model
  attr_accessor :login, :password
  validates :login, :password, presence: true
end

# domains/users/services/sign_up.rb
class Users::SignUp
  def call(params)
    form = Users::SignupForm.new(params)
    if form.valid?
      # ...
    else 
      # ...
    end
  end
end
```

If push directory `domains` to autoload paths of either Rails or Zeitwerk, it have to define class `Users::Services::SignUp` instead of `Users::SignUp`, this leads me to read Zeitwerk code, then found `Loader#set_autoloads_in_dir` already be designed to pass module/class other than `Object`. So, I can easily add a feature for Loader, support auto-loading from specific module/class:

```ruby
Users = Module.new
loader = Zeitwerk::Loader.new
Dir.glob('domains/users/*').each do |dir|
  loader.push_dir dir, parent: Users if File.directory?(dir)
end
loader.setup
```

if a module is autoloaded from another loader, Symbol or Proc is helpful.

```ruby
# app/domains/users.rb
module Users
end

# app/domains/payments.rb
module Payments
end

loader = Zeitwerk::Loader.new
loader.push_dir 'app/domains'
loader.setup

loader = Zeitwerk::Loader.new
Dir.glob('domains/users/*').each do |dir|
  loader.push_dir dir, parent: :Users if File.directory?(dir)
end
loader.push_dir 'domains/payments', parent: -> { Payments } 
loader.setup
```

I'm not sure whether Zeitwerk is planing to add this feature to `#push_dir`, but I think it's helpful to my project  😄

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -174,13 +174,14 @@ module Zeitwerk
     # or descendants.
     #
     # @param path [<String, Pathname>]
+    # @param options [Hash]
     # @raise [Zeitwerk::Error]
     # @return [void]
-    def push_dir(path)
+    def push_dir(path, options = {})
       abspath = File.expand_path(path)
       if dir?(abspath)
         raise_if_conflicting_directory(abspath)
-        root_dirs[abspath] = true
+        root_dirs[abspath] = { parent: Object }.merge(options)
       else
         raise Error, "the root directory #{abspath} does not exist"
       end
@@ -240,7 +241,7 @@ module Zeitwerk
       mutex.synchronize do
         break if @setup
 
-        actual_root_dirs.each { |root_dir| set_autoloads_in_dir(root_dir, Object) }
+        actual_root_dirs.each { |root_dir| set_autoloads_in_dir(root_dir, parent_for_root_dir(root_dir)) }
         do_preload
 
         @setup = true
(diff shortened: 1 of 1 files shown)


fxn:
I don't really like this particular proposal.

Technically, I believe it would be doable to implement skipping directories. Like, saying to the library that `domains/users/forms` is not a namespace, you should descend into it right away and define autoloads into `Users`.

But I am reluctant to do so by now, and the ramifications, edge cases, new expectations that would follow after opening the contract in that direction are difficult to foresee.

Right now, projects managed by Zeitwerk have a well-defined, predictable, uniform structure. I like that. It is also the same kind of structure you have in so many other languages like Java, Elixir, or Perl, to name a few.

In your case, `Users::Services::SignUp` and `Users::Forms::SignUp` look like good choices, because the reader of the code see the organization right there in the name. If I remember correctly, Trailblazer also adopted this organization/naming convention to play well with (`classic`) Rails autoloading.

So, thanks for the patch, but by now I prefer not to change the requirements about the project structure.

xiaohui-zhangxh:
I was struggling between `Users::Services::SignUp` and `Users::SignUp`, thank to point out me which is better. Thanks.

PR #95, opened 2019-10-31
Raise a Zeitwerk::NameError when a file does not define the expected constant

casperisfine:
### Use case

This is related to https://github.com/rails/rails/pull/37599, in short I'm trying to remove all `require_dependency` usage in Rails, and the last one is here: https://github.com/rails/rails/blob/cb740af5c96ebe0beea01fb8c62e93fb581ad454/actionpack/lib/abstract_controller/helpers.rb#L154

What Rails is doing there is that it `require_dependency` the expected file to know wether the constant doesn't exist, or wether it's loading fails.

So rewrite that code for pure Zeitwerk usage, I need to be able to do the same differentiation, e.g. if I get a `NameError` trying to access `FooHelper`, it can be either `NameError: expected file app/helpers/foo_helper.rb to define constant FooHelper, but didn't` or `NameError: uninitialized constant FooHelper`.

If it's the former I should re-raise it, if it's the later I should ignore it.

@fxn 

cc @Edouard-Chin, @etiennebarrie @rafaelfranca

### DIFF
--- a/lib/zeitwerk/loader/callbacks.rb
+++ b/lib/zeitwerk/loader/callbacks.rb
@@ -14,7 +14,7 @@ module Zeitwerk::Loader::Callbacks
     if logger && cdef?(*cref)
       log("constant #{cpath(*cref)} loaded from file #{file}")
     elsif !cdef?(*cref)
-      raise NameError, "expected file #{file} to define constant #{cpath(*cref)}, but didn't"
+      raise Zeitwerk::NameError, "expected file #{file} to define constant #{cpath(*cref)}, but didn't"
     end
   end
 
--- a/test/lib/zeitwerk/test_exceptions.rb
+++ b/test/lib/zeitwerk/test_exceptions.rb
@@ -5,7 +5,7 @@ class TestExceptions < LoaderTest
     files = [["typo.rb", "TyPo = 1"]]
     with_setup(files) do
       typo_rb = File.realpath("typo.rb")
-      e = assert_raises(NameError) { Typo }
+      e = assert_raises(Zeitwerk::NameError) { Typo }
       assert_equal "expected file #{typo_rb} to define constant Typo, but didn't", e.message
     end
   end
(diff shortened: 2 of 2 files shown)


fxn:
Makes sense. Let me think a bit about the exception name.

How would the new code in Rails look like using this change?

fxn:
Now that I remember, the library raises `Zeitwerk::NameError` since [2.1.10](https://github.com/fxn/zeitwerk/blob/master/CHANGELOG.md#2110-6-september-2019). Would that be enough?

casperisfine:
> Now that I remember, the library raises `Zeitwerk::NameError` since [2.1.10](https://github.com/fxn/zeitwerk/blob/master/CHANGELOG.md#2110-6-september-2019). Would that be enough?

And I'm the one that PRed it 🤦‍♂ how did I miss that...

What's weird is that I'm using `2.2.0` and I get a standard `NameError`:

```
[1] pry(main)> 'MoneyHelper'.constantize rescue $!
=> #<NameError: expected file /Users/byroot/src/github.com/Shopify/shopify/components/platform/app/helpers/money_helper.rb to define constant MoneyHelper, but didn't>
```

I'll need to dig a bit more into this, I might be missing something.

As for the code, I'm thinking of:

```ruby
if Rails.autoloaders.zeitwerk_enabled?
  args.flatten.map! do |arg|
    case arg
    when String, Symbol
      file_name = "#{arg.to_s.underscore}_helper"
      mod_name = file_name.camelize
      begin
        mod_name.constantize
      rescue Zeitwerk::NameError
        raise
      rescue NameError
        raise AbstractController::Helpers::MissingHelperError.new(e, file_name)
      end
    when Module
      arg
    else
      raise ArgumentError, "helper must be a String, Symbol, or Module"
    end
  end
else
  # existing code
end
```

But it's not definitive.

fxn:
> What's weird is that I'm using 2.2.0 and I get a standard NameError

Yeah, `Zeitwerk::NameError` is raised in a different use case. Guess it makes sense to raise here too.

casperisfine:
Ok, so turns out I was based on a very outdated master, which explains how I missed the `Zeitwerk::NameError`. Must have been tired.

I modified the PR, it should be good to go now.

fxn:
Perfect!

fxn:
I have published 2.2.1, thanks Jean!

PR #99, opened 2019-11-29
Properly set Zeitwerk::NameError#name

casperisfine:
Context: https://github.com/rails/rails/issues/37650

TL;DR; vanilla `NameError` has a `#name` accessor that is useful to distinguish wether the constant we were trying to load is missing, or if the `NameError` is caused by some dependent code.

e.g. being able to tell the difference between:

```ruby
# foo_controller.rb
class FoosControler # The expected constant doesn't exist
end
```

```ruby
# foo_controller.rb
class FooControler
 ::DoesNotExist # Another constant resolution issue was triggered by loading foo_controlle.rb
end
```

What this patch does is to respect the `NameError#name` contract.

@fxn for review please.

cc @rafaelfranca @etiennebarrie @Edouard-Chin

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -494,7 +494,7 @@ module Zeitwerk
         rescue ::NameError => error
           path_type = ruby?(abspath) ? "file" : "directory"
 
-          raise NameError, <<~MESSAGE
+          raise NameError.new(<<~MESSAGE, error.name)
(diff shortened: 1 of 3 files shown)


fxn:
Thanks @casperisfine!

fxn:
2.2.2 is out with this change.

casperisfine:
❤️

PR #105, opened 2020-01-27
Micro optimization: `slice(0...-3)` is actually faster than `slice!(-3, 3)`

casperisfine:
Not a huge deal at all, but I noticed a surprising amount of allocations from this idiom:

```
Allocated String Report
    109381  ""
     13344  ~/gems/zeitwerk-2.2.2/lib/zeitwerk/loader.rb:479
```

Turns out `slice!` allocates a wierd empty string for no reasons,
as well as a string containing the sliced chars we don't care about:

```ruby
>> s = "foo.rb"; MemoryProfiler.report { s.slice!(-3, 3) }.pretty_print
Allocated String Report
         1  ""
         1  (irb):20

         1  ".rb"
         1  (irb):20
```

Based on the context, I assume the `slice!` usage goal was to save some
string allocations, but turns out it actually makes things worse.

Two other candidate methods could be:

  - `byteslice`, as it ignore encoding it's marginally faster than `slice`.
  - `delete_suffix!` saves an allocation, but was introduced in Ruby 2.5.

Benchmark:

```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.report('slice!(-3, 3)') { "foo.rb".slice!(-3, 3) }
  x.report('slice(0...-3)') { "foo.rb".slice(0...-3) }
  x.report('byteslice(0...-3)') { "foo.rb".byteslice(0...-3) }
  x.report('delete_suffix!') { "foo.rb".delete_suffix!(".rb".freeze) }
end
```

```
       slice!(-3, 3)      4.939M (± 1.0%) i/s -     24.900M in   5.042534s
       slice(0...-3)      7.323M (± 1.4%) i/s -     36.774M in   5.023076s
   byteslice(0...-3)      8.126M (± 1.5%) i/s -     40.930M in   5.038314s
      delete_suffix!     10.657M (± 5.8%) i/s -     53.283M in   5.025248s
```

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -477,7 +477,7 @@ module Zeitwerk
       ls(dir) do |basename, abspath|
         begin
           if ruby?(basename)
-            basename.slice!(-3, 3)
+            basename[-3..-1] = ''
             cname = inflector.camelize(basename, abspath).to_sym
             autoload_file(parent, cname, abspath)
           elsif dir?(abspath)


fxn:
Good one! 

I believe we could go with `byteslice` then, since the `ruby?` predicate is

```ruby
def ruby?(path)
  path.end_with?(".rb")
end
```

and guards that code.

casperisfine:
> I believe we could go with `byteslice`

Yes, it just seemed a bit weird, so I wasn't sure. I'll update the PR.

casperisfine:
Done.

fxn:
Maybe amend the commit message?

casperisfine:
Of course.

casperisfine:
BTW for posterity's sake, I got a fix for the extra `slice!` allocation: https://github.com/Shopify/ruby/commit/12838feadcfb4f394f3a93134cfbe76fb660f375

TL;DR; `slice!(range)` internally calls `self[range] = ''` which causes that extra allocation we see.

I'll see about submitting it upstream, but I'm really not sure such small perf fix would be accepted.

casperisfine:
Damn, wait a bit. While writing this I just realized that `basename[]= ''` could actually be faster. Let me update my benchmark.

casperisfine:
So on the micro benchmark it's very sligthly slower than byteslice
```
       slice!(-3, 3)      4.915M (± 1.3%) i/s -     24.821M in   5.050738s
   byteslice(0...-3)      8.125M (± 0.6%) i/s -     40.815M in   5.023349s
       [-3..-1] = ""      7.842M (± 0.7%) i/s -     39.249M in   5.005565s
      delete_suffix!     10.920M (± 0.5%) i/s -     54.864M in   5.024335s
```

However it has the advantage of being entirely allocation free:
```
>> s = "foo.rb"; MemoryProfiler.report { s[-3..-1] = ''.freeze }.pretty_print
Total allocated: 0 bytes (0 objects)
Total retained:  0 bytes (0 objects)
```

Up to you if you'd like to use that last alternative.

fxn:
Awesome, let's go with that one then!

casperisfine:
I updated the commit.

fxn:
Thanks Jean! ❤️

casperisfine:
Welcome. TBH it was more about investigating that weird allocation than actually making anything noticeably faster.

PR #123, opened 2020-05-27
Correct GemInflector comment

bf4:
GemInflector.new requires an argument, the root file

naming of `root_file`  made with reference to current implementation of the code

## background

my code broke from zeitwerk 2.1.4 to 2.1.5 with the change from `require` to `const_get`.

tracking it down, it turned out it's because I'm doing the following

```
rails_root
  app
  ...etc..
  company_name/lib/company_name.rb
  company_name-integration_name/lib/company_name/integration_name.rb
```

i.e. there's a 'company' gem and various integrations are under that namespace

so, we have a Gemfile with something like

```ruby
  gem 'company_name', path: '.' # require: 'company_name' 
  gem 'company_name-integration_name', path: '.' # require 'company_name/integration_name`
```

the problem is that using the GemInflector now tried to `const_get` IntegrationName under `Object` instead of under `CompanyName`.  

My resolution of that is in 'company_name/integration_name/lib/company_name/integration_name.rb' to reuse the `for_gem` but set the loader one directory up

```ruby
require 'company_name'
require 'company_name/integration_name/version'
require 'company_name/integration_name/rails'
require 'zeitwerk'
loader_root_file = __FILE__
# needs to set the 'root' to 'company_name' not 'company_name/integration_name' so that require 'company_name-integration_name'
# looks up the 'IntegrationName' constant in under the object 'company_name' and not 'Object'
root_file = File.expand_path('..', __FILE__)
loader = Zeitwerk::Registry.loaders_managing_gems[root_file] ||=
  begin
    Zeitwerk::Loader.new.tap do |zl|
      zl.tag = File.basename(loader_root_file, ".rb")
      zl.inflector = Zeitwerk::GemInflector.new(loader_root_file)
      zl.push_dir(File.dirname(root_file))
    end
end
loader.ignore("#{__dir__}/integration_name/rails.rb") # defines an Engine, not Rails
loader.inflector.inflect(
  'company_name' => 'CompanyNAME',
)
module CompanyNAME
  module IntegrationName
  end
end
loader.eager_load
```
```

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -463,9 +463,10 @@ module Zeitwerk
       #
       #   require "zeitwerk"
       #   loader = Zeitwerk::Loader.new
-      #   loader.tag = File.basename(__FILE__, ".rb")
-      #   loader.inflector = Zeitwerk::GemInflector.new
-      #   loader.push_dir(__dir__)
+      #   root_file = __FILE__
+      #   loader.tag = File.basename(root_file, ".rb")
+      #   loader.inflector = Zeitwerk::GemInflector.new(root_file)
+      #   loader.push_dir(File.dirname(root_file))
       #
       # except that this method returns the same object in subsequent calls from
       # the same file, in the unlikely case the gem wants to be able to reload.

### CODE REVIEW

**Comment by fxn** [comment]
Thanks for the edit! That comment needs indeed to be fixed.

However, I have the feeling that your setup could be simplified. Why do you set path to `.` instead of the corresponding subdirectories?

```ruby
gem 'company_name', path: 'company_name' # require: 'company_name' 
gem 'company_name-integration_name', path: 'company_name/integration_name'
```

That would the normal thing to configure. It's a bit unusual to have a gem in a subdirectory of another gem, but technically should work because it is not below `lib`.

With that, loaders would be normal. The only important bit is that the owner of the `CompanyName` namespace is the `company_name` gem. Therefore, the other one should

```ruby
require "company_name"
```

at the top of the root file to let the main gem _define_ the namespace. That way the loader in the secondary one knows the `CompanyName` namespace is _redefined_ in its gem and should not attempt to set autoloads for it.

With that, the setup would be standard, and in principle you should be able to use vanilla `for_gem` calls in both gems.

fxn:
The README has more details about how to reopen namespaces, see [this section](https://github.com/fxn/zeitwerk#reopening-third-party-namespaces).

bf4:
@fxn Thanks for the quick response.  I had a typo in my description and may have otherwise been unclear.

image I typed `rails new my_app` then cd into my_app and that is 'rails_root' above.  Then I typed `rails g plugin company_name` then `rails g plugin company_name-integration_name`.  i.e. the two gems are internal to the Rails app and both in the Rails root

- description before <br> Screen Shot 2020-05-27 at 5 04 56 PM

- description after <br>  Screen Shot 2020-05-27 at 5 05 32 PM

bf4:
The namespaces section *does* describe my scenario but didn't work. I'm pretty sure my folder structure is right.

bf4:
@fxn I can make a general issue around my situation if you'd like or keep discussing here or elsewhere.

fxn:
Could you please use `__FILE__` in the constructor and keep `__dir__`?

fxn:
While the gems live in the Rails project root, Rails ignores them. In Rails, `Rails.root` is not in `$LOAD_PATH` or in autoload paths.

The Gemfile should treat them as independent gems and they should have `path` set accordingly. The `path` option points to the root directory of a gem, so:

```ruby
gem 'company_name', path: 'company_name'
gem 'company_name-integration_name', path: 'company_name-integration_name'
```

Idiomatically, indeed, they would go into `vendor` better than the root:

```ruby
gem 'company_name', path: 'vendor/company_name'
gem 'company_name-integration_name', path: 'vendor/company_name-integration_name'
```

If you do that, Zeitwerk setup would be the usual one-liner.

fxn:
Hey Benjamin. I have generated those plugins. For the second one, this should be working starting point:

```ruby
require "company_name"
require "company_name/integration_name/railtie"

require "zeitwerk"
loader = Zeitwerk::Loader.new
loader.tag = "company_name-integration_name"
loader.push_dir(File.expand_path("..", __dir__))
loader.ignore(File.expand_path("../tasks", __dir__))
loader.inflector.inflect("version" => "VERSION")
loader.setup

module CompanyName
  module IntegrationName
  end
end
```

The method `for_gem` is only a shortcut for the common case in which the root file is directly under `lib`. But this one does not satisfy that. In this case, you need to setup the loader by hand. 

Does that help? Remember to edit `path`!

lib/zeitwerk/loader.rb
```diff
@@ -463,9 +463,10 @@ class << self
       #
       #   require "zeitwerk"
       #   loader = Zeitwerk::Loader.new
-      #   loader.tag = File.basename(__FILE__, ".rb")
-      #   loader.inflector = Zeitwerk::GemInflector.new
-      #   loader.push_dir(__dir__)
+      #   root_file = __FILE__
+      #   loader.tag = File.basename(root_file, ".rb")
+      #   loader.inflector = Zeitwerk::GemInflector.new(root_file)
+      #   loader.push_dir(File.dirname(root_file))
```
bf4, line 469: @fxn  you wrote:
> Could you please use `__FILE__` in the constructor and keep `__dir__`?

sure. I wrote it that way because that's how it's used internally and seemed to me reasonable to align them in this way

fxn:
Hello Benjamin!

I like to keep the counters of this project down to 0, so I pushed the doc fix myself giving you credit (https://github.com/fxn/zeitwerk/commit/8adfb118df1e7df8f837ca0c51253bbdf50d1a54).

On the other hand, the documentation of `for_gem` in the README now explictly addresses your use case, see the third paragraph of [this section](https://github.com/fxn/zeitwerk#for_gem).

Finally, I strongly encourage you to check the advices above. For simplicity, and to avoid using methods which are marked as private. Such methods can work in a different way or just disappear in any of the future releases, no matter how tiny the version increment.

Thanks for the patch and for sharing your use case, hope to meet you again in some conference!

bf4:
Thanks @fxn 

I took all your advice and things are working fine.  I think I must have been doing paths in gemfiles like that for 10 years. shrug

### all the internal gems are now in vendor

And the Gemfile in our Rails app is

```ruby
# Our Gems
path 'vendor' do
  gem "company_name", path: 'vendor/company_name'
  gem "company_name-integration_name", path: 'vendor/company_name-integration_name' # depends on company_name
end
```


### the 'company_name' gem is now

in vendor/company_name/lib/company_name.rb

```ruby
require "zeitwerk"
loader.inflector.inflect(
  'company_name' => 'CompanyNAME',
)
loader.setup

module CompanyNAME
  # @example
  # loader = CompanyNAME.setup_loader(
  #   file: __FILE__,
  #   dir: __dir__,
  # )
  # loader.ignore("#{__dir__}/integration_name/rails.rb")
  # loader.setup
  #
  # @note for gems under company_name/
  #   needs to set the 'root' to 'company_name' not 'company_name/integration_name' so that require 'company_name/integration-name'
  #   looks up the 'IntegrationName' constant in under the object 'CompanyNAME' and not 'Object'
  #   and see https://github.com/fxn/zeitwerk/pull/123
  def self.setup_autoloader(file:, dir:)
    loader = Zeitwerk::Loader.new
    loader.push_dir(File.expand_path("..", dir))
    loader.tag = File.basename(file, ".rb")
    loader.ignore(File.expand_path("../tasks", dir))
    loader.inflector.inflect(
      'company_name' => 'CompanyNAME',
      "version" => "VERSION"
    )
    loader
  end
end

loader.eager_load
```

### the 'company_name-integration' gem is now

in vendor/company_name-integration_name/lib/company_name/integration_name.rb

```ruby
require "company_name"
require "company_name/integration_name/rails"
require 'zeitwerk'

loader = CompanyNAME.setup_autoloader(
  file: __FILE__,
  dir: __dir__,
)
loader.ignore("#{__dir__}/integration_name/rails.rb")
loader.setup

module CompanyNAME
  module IntegrationName
  end
end

loader.eager_load
```

PR #125, opened 2020-06-28
Freeze absolute paths for reduce allocations on file operations

casperisfine:
Most of the `File` methods cast the paths it receives with [`FilePathValue / rb_get_path`](https://github.com/ruby/ruby/blob/09b936d89cb66e38db46dbe782aa5f22f94656bc/file.c#L244-L248) which ultimately delegate to [`rb_new_str_frozen`](https://github.com/ruby/ruby/blob/09b936d89cb66e38db46dbe782aa5f22f94656bc/file.c#L225-L236) (aliased as `rb_str_new4`).

[`rb_new_str_frozen` could be experessed as `str.frozen? ? str : str.dup`.](https://github.com/ruby/ruby/blob/09b936d89cb66e38db46dbe782aa5f22f94656bc/string.c#L1217-L1222)

So by freezing the paths passed to the various `File.` methods, we
save at least one useless allocation each time, unless the string encoding
is incompatible with the file system.

```
s = __FILE__
alloc = GC.stat[:total_allocated_objects]
File.realpath(s)
p GC.stat[:total_allocated_objects] - alloc # => 3

s = __FILE__.freeze
alloc = GC.stat[:total_allocated_objects]
File.realpath(s)
p GC.stat[:total_allocated_objects] - alloc # => 2
```

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -616,7 +616,7 @@ module Zeitwerk
       # $LOADED_FEATURES stores real paths since Ruby 2.4.4. We set and save the
       # real path to be able to delete it from $LOADED_FEATURES on unload, and to
       # be able to do a lookup later in Kernel#require for manual require calls.
-      realpath = File.realpath(abspath)
+      realpath = File.realpath(abspath).freeze
       parent.autoload(cname, realpath)
       if logger
         if ruby?(realpath)
@@ -719,7 +719,7 @@ module Zeitwerk
     def ls(dir)
       Dir.foreach(dir) do |basename|
         next if basename.start_with?(".")
-        abspath = File.join(dir, basename)
+        abspath = File.join(dir, basename).freeze
         yield basename, abspath unless ignored_paths.member?(abspath)
       end
     end


lib/zeitwerk/loader.rb
```diff
@@ -377,7 +377,7 @@ def eager_load
             next if eager_load_exclusions.member?(abspath)
 
             if ruby?(abspath)
-              if cref = autoloads[File.realpath(abspath)]
+              if cref = autoloads[File.realpath(abspath).freeze]
```
fxn, line 380: Which is the point of this one?

fxn:
I've checked with `autoload` too. It saves three allocations!

```
s = __FILE__
alloc = GC.stat[:total_allocated_objects]
Object.autoload(:A, s)
p GC.stat[:total_allocated_objects] - alloc # => 5

s = __FILE__.freeze
alloc = GC.stat[:total_allocated_objects]
Object.autoload(:B, s)
p GC.stat[:total_allocated_objects] - alloc # => 2
```

lib/zeitwerk/loader.rb
casperisfine: It's just a brainfart. I'll remove it.

casperisfine:
> I've checked with `Module#autoload` too. It saves three allocations!

Yeah, `autoload` calls `realpath`, which itself allocates a bit too much. I'm currently working on upstream PRs to improve that (https://github.com/ruby/ruby/pull/3267).

But it seems like all File operation always allocate at least one object:

```ruby
s = __FILE__.freeze
alloc = GC.stat[:total_allocated_objects]
File.exist?(s)
p GC.stat[:total_allocated_objects] - alloc # => 1
```

It's not allocating a String though, I'm trying to pinpoint that alloc. Anyway it's a bit unrelated.

fxn:
Perfect!

casperisfine:
> But it seems like all File operation always allocate at least one object:

Ok so self answer, even calling an empty method cause 1 alloc, so 1 is the minimum.

PR #126, opened 2020-06-29
Use Dir.each_child rather than Dir.foreach

casperisfine:
`Dir.each_child` don't yield `.` and `..` which we don't care about.

That's an extremely minor improvement, but it took me literally 5 seconds.


fxn:
IIRC, I tried at some point in the past with the same motivation, but if my memory is correct it was not available in Ruby 2.4 which the gem still supports (though it reached EOL). I can confirm/refute later.

casperisfine:
Looks like you're right: https://ruby-doc.org/core-2.4.0/Dir.html#method-c-each_child

My bad.

PR #132, opened 2020-08-21
Add truffleruby-head to CI

bjfish:
Please add truffleruby-head to travis CI so we can see the status of the latest TruffleRuby.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ rvm:
   - ruby-head
   - jruby-head
   - truffleruby
+  - truffleruby-head
 matrix:
   allow_failures:
     - rvm: jruby-head
+    - rvm: truffleruby-head
     - rvm: truffleruby


fxn:
In case it helps, Zeitwerk mutates `$LOADED_FEATURES` using `Array#reject!` [here](https://github.com/fxn/zeitwerk/blob/1790e82c579e118522c181e37c3f4870e17deb56/lib/zeitwerk/loader.rb#L334).

The motivation for that is that reloading is implemented as "remove constant + load its definition again". Since `Module#autoload` loads with `require`, we need to mutate `$LOADED_FEATURES` to workaround its idempotency.

bjfish:
@fxn We found a bug related to the test failures and a fix has been pushed: https://github.com/oracle/truffleruby/commit/985b21c80c1dcb4bf41e8cde28aefd0a10ca107c. I expect future builds to pass for `truffleruby-head`.

fxn:
@bjfish I am curious, why the emphasis on `pop`/`shift`? The API to mutate arrays has several methods, and you could combine them to edit the thing in place at arbitrary positions and preserve the size.

eregon:
Other mutations will cause a copy of the the underlying storage (which is copy-on-write).
`pop`/`shift` do not need to modify the underlying storage, and so they just adjust the size or offset.

eregon:
@fxn Assuming `truffleruby-head` passes the CI reliably now, do you think we could remove it from `allow_failures`?

fxn:
@eregon [Removed](https://github.com/fxn/zeitwerk/commit/627482af33c895c7199b261e78f4d8037b01938e) and [green](https://travis-ci.com/github/fxn/zeitwerk) 🎉.

Issue #129, opened 2020-07-11
Evaluate opening a bit the way to define explicit namespaces


fxn:
## Objective

Some people have asked for other ways to define explicit namespaces. The purpose of this issue is to put together known use cases/feature requests and see if something well aligned with the library is possible.

## Use Case 1

In #124, there's a Sinatra application whose desired layout would be

```
root/
  appname/
    components/
      homepage/
        homepage.rb         => AppName::Homepage
        homepage.html.erb
        homepage.js
        welcome_box/
          welcome_box.rb    => AppName::Homepage::WelcomeBox
```

There, `components` is a collapsed directory.

## Use Case 2

Interestingly, in #128 the use case mentions the word "component" too. There, the feature request is to support one special file in the namespace directory with an arbitrary but fixed name, like `index.rb`:

```
root/
  component/
    index.rb => Component
    foo.rb   => Component::Foo
```

## Use Case 3

In [this post](https://discuss.rubyonrails.org/t/tales-from-an-unconventional-rails-app/75250/3), the missed feature is to be able to define the namespace `FooBar` in

```
projects/foo_bar/config/project.rb
```

## Benefits

* Perceived better code organization in some particular use cases.
* Components can be implemented as Git submodules.
* Easier navigation in editors.

fxn:
@marcrohloff would you mind giving a real example with real names for `component` and `foo.rb`? Is yours a Rails application? 

@pvande would you mind detailing the layout of that project and use a real name for `foo_bar`? Is `config` a root directory? Is yours a Rails application?

pvande:
@fxn Here’s an example of our project’s file structure:

```
/our_rails_app
├── Gemfile
├── Gemfile.lock
├── app
│   ├── assets
│   ├── controllers
│   ├── helpers
│   ├── models
│   └── views
├── bin
├── config
├── db
├── lib
├── log
├── projects
│   ├── k12_science
│   │   ├── assets
│   │   ├── config
│   │   ├── forms
│   │   ├── helpers
│   │   ├── lib
│   │   ├── models
│   │   ├── services
│   │   ├── tasks
│   │   ├── test
│   │   └── views
│   └── k12_math
│       ├── assets
│       ├── config
│       │   ├── locales
│       │   ├── postcss.config.js
│       │   ├── project.rb      (defines `K12Math`)
│       │   └── webpack.js
│       ├── forms
│       │   ├── unit_form.rb    (defines `K12Math::UnitForm`)
[...]

PR #137, opened 2020-10-03
Spike with codes enabling to be defined within their own component folder

marcrohloff:
wrt to this discussion [#129: Evaluate opening a bit the way to define explicit namespaces](https://github.com/fxn/zeitwerk/issues/129). I had a bit of time to create an experimental branch and see what would be required. This is what I cam up with.

To address concerns
* it is opt in
* Does not cause extra File stat's when not enabled
* Is flexible (The index_filespec can be almost anything `index` `~index`, `*`, `~*`)

I tested with the following model tree
├── app
│   ├── models
│   │   ├── components
│   │   │   ├── hotel
│   │   │   │   └── room.rb
│   │   │   ├── house
│   │   │   │   └── room.rb
│   │   │   ├── house.rb
│   │   │   └── office
│   │   │       ├── ~office.rb
│   │   │       └── room.rb

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -18,6 +18,26 @@ module Zeitwerk
     # @return [#call, #debug, nil]
     attr_accessor :logger
 
+    # Enable loading modules using index files in addition to searching the normal location
+    #
+    # The pattern for the file you want to use for an index file ('index', '~index', '*', '*_module')
+    #
+    # Setting a value will cause ZeitWerk to look for a module under the given name in a module sub-folder
+    # in addition to the normal location. i.e. When searching for the `Hello` module it will look in hello/index.rb
+    # in addition to hello.rb.
+    #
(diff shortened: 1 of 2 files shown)


fxn:
Thanks for the patch. The ticket was closed with a resolution, but I'll consider this anyway. Looking up alternative file in `on_dir_autoloaded` is clever.

Let me give it some time.

fxn:
There is a technical issue: Engines use the same autoloader as their parent application. If you ship an engine as open source with a file `foo/index.rb` that defines `Foo::Index` and the parent application configures Zeitwerk to expect the base module from there, the engine is broken. In the future, maybe we manage to refactor all this so that each engine has its own autoloader, but by now a global setting like that would have that consequence.

However, even if that technical issue did not exist, and after sleeping on it a few nights, [the rationale](https://github.com/fxn/zeitwerk/issues/129#issuecomment-660513782) for closing the original issue still holds.

Thanks for working on this anyway!

PR #144, opened 2020-11-26
Implements Zeitwerk::Loader#on_load

### LINKED ISSUE #142: Feature Request: Zeitwerk.on_load
### Context

It is very common in Rails apps, to set configuration values on module / classes from either `config/environments/*.rb` or `config/initializers/*.rb`. e.g.:

```ruby
# some/autoloaded/file.rb
class SomeApiClient
  class << self
    attr_accessor :endpoint
  end
end
```

```ruby
# config/environments/development.rb
config.to_prepare do
  SomeApiClient.endpoint = 'https://my-api.dev'
end
```

```ruby
# config/environments/production.rb
config.to_prepare do
  SomeApiClient.endpoint = 'https://my-api.com'
end
```

### Problem

Because of this pattern, a lot of code that should be lazy loaded, end up eager loaded. Assuming part of this code won't be used in the development request, or test case, it means time is wasted loading "useless" code.

### Proposal

I think `zeitwerk` could offer an API to get a callback when a namespace is loaded, [similar to `ActiveSupport::LazyLoadHooks`](https://api.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html), e.g.:

```ruby
# config/environments/development.rb
Zeitwerk.on_load('SomeApiClient') do |client|
  client.endpoint = 'https://my-api.dev'
end

# or

Zeitwerk.on_load('SomeApiClient') do
  SomeApiClient.endpoint = 'https://my-api.dev'
end

# or

Zeitwerk.on_load('SomeApiClient') do
  self.endpoint = 'https://my-api.dev'
end
```

@fxn what do you think?

cc @rafaelfranca @etiennebarrie

fxn:
Title says it all, see the documentation in this patch for details.

Closes #142.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@
         - [Zeitwerk::Inflector](#zeitwerkinflector)
         - [Zeitwerk::GemInflector](#zeitwerkgeminflector)
         - [Custom inflector](#custom-inflector)
+    - [The on_load callback](#the-on_load-callback)
     - [Logging](#logging)
         - [Loader tag](#loader-tag)
     - [Ignoring parts of the project](#ignoring-parts-of-the-project)
@@ -501,6 +502,52 @@ class MyGem::Inflector < Zeitwerk::GemInflector
   # ...
 end
 ```
+
+<a id="markdown-the-on_load-callback" name="the-on_load-callback"></a>
+### The on_load callback
+
+The usual place to run something when a file is loaded is the file itself. However, sometimes you'd like to be called, and this is possible with the `on_load` callback.
+
+For example, let's imagine this class belongs to a Rails application:
+
+```ruby
+class SomeApiClient
+  class << self
+    attr_accessor :endpoint
+  end
+end
+```
+
+With `on_load`, it is easy to schedule code at boot time that initializes `endpoint` according to the configuration:
+
+```ruby
+# config/environments/development.rb
+loader.on_load("SomeApiClient") do
+  SomeApiClient.endpoint = "https://api.dev"
+end
+
+# config/environments/production.rb
+loader.on_load("SomeApiClient") do
+  SomeApiClient.endpoint = "https://api.prod"
+end
+```
+
+Uses cases:
+
+* Doing something with an autoloadable class or module in a Rails application during initialization, in a way that plays well with reloading. As in the previous example.
+* Delaying the execution of the block until the class is loaded for performance.
+* Delaying the execution of the block until the class is loaded because it follows the adapter pattern and better not to load the class if the user does not need it.
+* Etc.
+
+However, let me stress that the easiest way to accomplish that is to write whatever you have to do in the actual target file. `on_load` use cases are edgy, use it only if appropriate.
+
+`on_load` receives the name of the target class or module as a string. The given block is executed every time its corresponding file is loaded. That includes reloads.
+
+Multiple callbacks on the same target are supported, and they run in order of definition.
+
+The block is executed once the loader has loaded the target. In particular, if the target was already loaded when the callback is defined, the block won't run. But if you reload and load the target again, then it will. Normally, you'll want to define `on_load` callbacks before `setup`.
+
+Defining a callback for a target not managed by the receiver is not an error, the block simply won't ever be executed.
 
 <a id="markdown-logging" name="logging"></a>
 ### Logging
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -129,6 +129,9 @@ module Zeitwerk
     # @sig Set[String]
     attr_reader :eager_load_exclusions
 
+    # User-oriented callbacks to be fired when a constant is loaded.
+    attr_reader :on_load_callbacks
+
     # @private
     # @sig Mutex
     attr_reader :mutex
@@ -155,6 +158,7 @@ module Zeitwerk
       @to_unload              = {}
       @lazy_subdirs           = {}
       @eager_load_exclusions  = Set.new
+      @on_load_callbacks      = {}
 
       # TODO: find a better name for these mutexes.
       @mutex        = Mutex.new
@@ -259,6 +263,24 @@ module Zeitwerk
       mutex.synchronize do
         collapse_glob_patterns.merge(glob_patterns)
         collapse_dirs.merge(expand_glob_patterns(glob_patterns))
+      end
+    end
+
+    # Configure a block to be invoked once a certain constant path is loaded.
+    # Supports multiple callbacks, and if there are many, they are executed in
+    # the order in which they were defined.
+    #
+    #   loader.on_load("SomeApiClient") do
+    #     SomeApiClient.endpoint = "https://api.dev"
+    #   end
+    #
+    # @raise [TypeError]
+    # @sig (String) { () -> void } -> void
+    def on_load(cpath, &block)
+      raise TypeError, "on_load only accepts strings" unless cpath.is_a?(String)
+
+      mutex.synchronize do
+        (on_load_callbacks[cpath] ||= []) << block
       end
     end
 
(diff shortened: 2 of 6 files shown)


lib/zeitwerk/loader.rb
```diff
@@ -262,6 +266,23 @@ def collapse(*glob_patterns)
...
+    # Supports multiple callbacks, and if there are many, they are executed in
+    # the order in which they were defined.
+    #
+    #   loader.on_load("SomeApiClient") do
+    #     SomeApiClient.endpoint = "https://my-api.dev"
+    #   end
+    #
+    # @sig (String) { () -> void } -> void
+    def on_load(cpath, &block)
+      raise Error, "on_load only accepts strings" unless cpath.is_a?(String)
```
casperisfine, line 279: Might make sense to raise a `::TypeError` here.

lib/zeitwerk/loader.rb
```diff
@@ -262,6 +266,23 @@ def collapse(*glob_patterns)
...
+
+      mutex.synchronize do
+        (on_load_callbacks[cpath] ||= []) << block
```
casperisfine, line 282: You didn't seem so keen on it, but shouldn't we invoke the callback immediately if the module or class is already loaded? Or is there some adverse effects to that I'm not thinking of?

fxn: It is a matter of how we want to define this feature. If you define an event on a button push, it does not matter if the button was pushed before.

In our case, furthermore, we do not know what we autoloaded if eager loading. A `const_defined?` is not scoped to the loader.

All in all, I think this logic is more clear, and enough for the use cases I have in mind, which happen before `setup`.

lib/zeitwerk/loader.rb
fxn: Good point.

README.md
```diff
@@ -502,6 +503,43 @@ class MyGem::Inflector < Zeitwerk::GemInflector
...
+With `on_load`, it is easy to initialize said attribute at boot time according to the configuration:
+
+```ruby
+# config/environments/development.rb
+loader.on_load("SomeApiClient") do
+  SomeApiClient.endpoint = "https://api.dev"
+end
+
+# config/environments/production.rb
+config.on_load("SomeApiClient") do
```
etiennebarrie, line 530: It should be `loader` here right? Also in the changelog.

lib/zeitwerk/loader.rb
casperisfine: > It is a matter of how we want to define this feature. If you define an event on a button push, it does not matter if the button was pushed before.

Agreed. But my definition of this feature is: "Execute this block as soon as this constant is loaded". Which would be consistent with `ActiveSupport.on_load`: https://api.rubyonrails.org/classes/ActiveSupport/LazyLoadHooks.html#method-i-on_load

 > In our case, furthermore, we do not know what we autoloaded if eager loading

I think we could just execute directly if eager loading.

Either way I can work around this limitation myself by using a higher level help, but I do really think that's what users expcetions and need would be.

README.md
```diff
@@ -502,6 +503,43 @@ class MyGem::Inflector < Zeitwerk::GemInflector
...
+end
+```
+
+This method receives the name of the target class or module as a string. The given block is executed every time its corresponding file is loaded. That includes reloads.
+
+Multiple callbacks on the same target are supported, and they run in order of definition.
+
+Think about `on_load` as reacting to an event. In particular, if the target was already loaded when the callback is defined, the block won't run. Normally, these callbacks are set before `setup`, but this is not enforced.
+
+Defining a callback for a target not managed by the receiver is not an error, the block simply won't be ever executed.
```
etiennebarrie, line 541: Not a native speaker but I think "won't ever be executed" is more idiomatic.

README.md
```diff
@@ -502,6 +503,43 @@ class MyGem::Inflector < Zeitwerk::GemInflector
...
+
+```ruby
+class SomeApiClient
+  class << self
+    attr_accessor :endpoint
+  end
+end
+```
+
+With `on_load`, it is easy to initialize said attribute at boot time according to the configuration:
```
etiennebarrie, line 521: Maybe we could mention somewhere that it's especially useful because it avoids loading a constant if you didn't need to?

etiennebarrie (commented):
Tiny docs feedback.

README.md
fxn: Good catch, it was introduced to see if reviewers paid attention 🤣.

casperisfine: I have to admit I never review docs cause I'm so bad at it...

README.md
fxn: Good, I'll add something in that line. It also a way to avoid autoloading during boot time in Rails, and in a way that works well with reloading.

I'll give it a thought.

lib/zeitwerk/loader.rb
fxn: I guessed you had `AS::.on_load` in mind, but parity with that interface is not a goal of this feature per se. I need to think about what makes sense or is possible in this gem, and what does a callback mean in general. There are many callbacks that are only reactive, in JavaScript, in Active Record, in controllers, etc., so a reactive callback is a valid interface.

When I wrote "eager loading" I mean "reloading disabled", corrected it later, but too late for you to read it again.

In which use case you need to define a callback after `setup` and after the class has been loaded?

casperisfine: > In which use case you need to define a callback after `setup` and after the class has been loaded?

Oh right I think we already had that discussion. I suppose that if we only allow to define these during setup, it's not as much as a big deal.

I'm just very worried about some callback accidentally not being fired anymore because of a subtle change in load order. But maybe I'm being too paranoid.

casperisfine: Would it make sense to you to raise an error if someone tries to add a callback after the setup phase?

fxn: I thought about that, but in this case I leant on being permissive. By not raising, we allow callbacks to be defined on reload according to file changes. Not that I have a use case in mind, just a gut feeling of not being rigid where I do not see an obvious reason to be so.

There's a bit of intuition in this decision and it is debatable, someone else (maybe you :), could perfectly decide to raise.

In the end, these trade-offs end up in the documentation/spec of the feature. The README tells you how it works, and it makes explicit that you can define after `setup`, it is your responsibility to know what you are doing.

fxn: Ah, let me add that this flexibility has in mind that Zeitwerk is intended to be used by arbitrary Ruby projects, so Rails mental models should not constrain the design.

It is totally debatable, though.

casperisfine: Fine by me. I guess we'll see how it goes anyway.

casperisfine (approved):

lib/zeitwerk/loader.rb
fxn: Cool man, indeed it is tentative, time can perfectly prove it wrong.

README.md
fxn: I've written this:

> Uses cases:
> 
> * Doing something with an autoloadable class or module in a Rails application during initialization, in a way that plays well with reloading.
> * Delaying the execution of the block until the class is loaded for performance.
> * Delaying the execution of the block until the class is loaded because it follows the adapter pattern and better not to load the class if the user does not need it.
> * Etc.
> 
> However, let me stress that the easiest way to accomplish that is to put the code in the actual file. `on_load` is useful when from a design point of view it is better that the code in the callback is elsewhere.
> 

Going to wrap up and publish a new version soon.

README.md
fxn: Confirmed that yours is the good one with a native speaker 👍.

fxn:
Version 2.4.2 is out including this, thanks! ❤️

casperisfine:
❤️ 

I'll only be able to integrate it by mid next week, but I'll happily report some feedback once I can if you are interested.

fxn:
> I'll happily report some feedback once I can if you are interested.

Always! :)

PR #145, opened 2020-11-30
Fix CHANGELOG formatting

casperisfine:
The title says it all.


fxn:
LOL, some day I'll actually remember how to do this. Thanks Jean!

PR #147, opened 2021-01-01
Add Ruby 3.0 on CI matrix

bogdanvlviv:
Also, I'm wondering whether `truffleruby` has been added to "allow_failures"
category by mistake? Suppose we should add `truffleruby-head` instead.
Since tests for `truffleruby` are green for a couple of the last builds
https://github.com/fxn/zeitwerk/runs/1577391820

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ rvm:
   - 2.5
   - 2.6
   - 2.7
+  - 3.0
   - ruby-head
   - jruby-head
   - truffleruby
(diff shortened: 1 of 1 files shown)


bogdanvlviv:
Builds of ruby-head are failed because of https://github.com/seattlerb/minitest/pull/862

fxn:
Hey man!

I have added `ruby-head` to the allowed failures while `minitest` gets fixed, and removed `truffleruby`, which was there just as an overlook.

fxn:
Thanks!

PR #157, opened 2021-04-02
Allow to disable symlink resolution

casperisfine:
Zeiwerk end up calling `File.realpath` a lot to not trip on symlinsk, and that is a relatively expensive syscall, even more so on systems with slow file systems.

For advanced users who know their load paths don't contain any symlinks, it can be beneficial to only resolve the root directories.

In our app we spend `~1.7%` of boot time in `File.realpath`, even though I know for a fact that there isn't a single symlink.

```
$ stackprof ~/Downloads/stackprof-shopify-boot-production-cpu.dump --method=File.realpath
File.realpath (<cfunc>:1)
  samples:   757 self (1.7%)  /    758 total (1.7%)
  callers:
     351  (   46.3%)  Zeitwerk::Loader#set_autoload
     349  (   46.0%)  Zeitwerk::Loader#eager_load
      38  (    5.0%)  Bootsnap::LoadPathCache::RealpathCache#find_file
       9  (    1.2%)  Bootsnap::CompileCache::YAML::Patch#load_file
       9  (    1.2%)  Bootsnap::LoadPathCache::Cache#initialize
       1  (    0.1%)  Rails::Engine.find_root_with_flag
       1  (    0.1%)  Google::Cloud.loaded_files
```

It would be nice it there was an "advanced" setting to turn this off.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -18,6 +18,9 @@ module Zeitwerk
     # @sig #call | #debug | nil
     attr_accessor :logger
 
+    # @sig Boolean
+    attr_accessor :resolve_symlinks
+
     # Absolute paths of the root directories. Stored in a hash to preserve
     # order, easily handle duplicates, and also be able to have a fast lookup,
     # needed for detecting nested paths.
@@ -159,6 +162,7 @@ module Zeitwerk
       @lazy_subdirs           = {}
       @eager_load_exclusions  = Set.new
       @on_load_callbacks      = {}
+      @resolve_symlinks       = true
 
       # TODO: find a better name for these mutexes.
       @mutex        = Mutex.new
@@ -202,11 +206,17 @@ module Zeitwerk
       end
 
       abspath = File.expand_path(path)
+      begin
+        abspath = File.realpath(abspath)
+      rescue Errno::ENOENT
+        raise Error, "the root directory #{abspath} does not exist"
+      end
+
       if dir?(abspath)
         raise_if_conflicting_directory(abspath)
         root_dirs[abspath] = namespace
       else
-        raise Error, "the root directory #{abspath} does not exist"
+        raise Error, "#{abspath} is not a directory"
       end
     end
 
@@ -405,7 +415,7 @@ module Zeitwerk
             next if eager_load_exclusions.member?(abspath)
 
             if ruby?(abspath)
-              if cref = autoloads[File.realpath(abspath)]
+              if cref = autoloads[realpath(abspath)]
                 cref[0].const_get(cref[1], false)
               end
             elsif dir?(abspath) && !root_dirs.key?(abspath)
@@ -632,7 +642,7 @@ module Zeitwerk
       #
       # We freeze realpath because that saves allocations in Module#autoload.
       # See #125.
-      realpath = File.realpath(abspath).freeze
+      realpath = realpath(abspath).freeze
       parent.autoload(cname, realpath)
       if logger
         if ruby?(realpath)
@@ -788,6 +798,15 @@ module Zeitwerk
     # @sig (String) -> void
     def register_explicit_namespace(cpath)
       ExplicitNamespace.register(cpath, self)
+    end
+
+    # @sig (String) -> String
+    def realpath(path)
+      if @resolve_symlinks
+        File.realpath(path)
+      else
+        path
+      end
     end
 
     # @sig (String) -> void
--- a/test/lib/zeitwerk/test_push_dir.rb
+++ b/test/lib/zeitwerk/test_push_dir.rb
@@ -38,4 +38,16 @@ class TesPushDir < LoaderTest
     e = assert_raises(Zeitwerk::Error) { loader.push_dir(".", namespace: :foo) }
     assert_equal ":foo is not a class or module object, should be", e.message
   end
+
+  test "resolve the real path even with resolve_symlinks = false" do
+    files = [["real/x.rb", "X = true"]]
+    with_files(files) do
+      FileUtils.ln_s("real", "symlink")
+      loader.resolve_symlinks = false
+      loader.push_dir("symlink")
+      loader.setup
+
+      assert_equal File.realpath("real/x.rb"), Object.autoload?(:X)
+    end
+  end
 end


fxn:
ACK! I'll give it a thought.

fxn:
I might need some time for this one. There are several trade-offs at play. In particular, if it weren't for the experimental and undocumented (on purpose) support for `require`, I believe Zeitwerk could work without `realpath` calls if reloading is disabled, which I consider to be the most common use case.

If we forget about `require` support, why do we need real paths? Mainly, to be able to maintain `$LOADED_FEATURES` (mod I might be misremembering some corner of the implementation, give me a margin of error). If support for `require` changes or is dropped, we should be able to unload just by keeping track of the real paths of autoloaded files. Not even for all autoloads set, just for those that have been executed. And if reloading is disabled, we should not need real paths at all, saving many calls.

For the archives, let me add that Ruby performs its own internal `realpath` calls in every single `require` or `require_relative`, to be able to check `$LOADED_FEATURES`. I wondered if the operating system cache for file stats makes the second of two close calls cheap, but benchmarks show the answer is negative:

```ruby
require "benchmark/ips"

ME = File.expand_path(__FILE__, __dir__)
$LOADED_FEATURES << File.realpath(ME)

Benchmark.ips do |x|
  x.report("realpath") do |n|
    i = 0
    while i < n
      File.realpath(ME)
      i += 1
    end
  end

  x.report("realpath + require") do |n|
    i = 0
    while i < n
      File.realpath(ME)
      require ME
      i += 1
    end
  end
end
```

```
% ruby foo.rb
Warming up --------------------------------------
            realpath     5.146k i/100ms
  realpath + require     1.388k i/100ms
Calculating -------------------------------------
            realpath     51.278k (± 3.4%) i/s -    257.300k in   5.023853s
  realpath + require     13.601k (± 2.1%) i/s -     68.012k in   5.002806s
```

This days I am dropping support for `classic` mode in `rails/main`, and I am waiting for that work to be complete to release 2.5 or 3.0, depending on the magnitude of the changes.

I'll give it time and see what we do.

casperisfine:
> I believe Zeitwerk could work without realpath calls if reloading is disabled, which I consider to be the most common use case.

Well, my profiles comes from booting the app in production mode, so reload disabled. However I think development mode (with reload enabled) need such optimization even more. Production mode is generally on Linux, whereas dev mode is very often on OSX. [And OSX has notoriously slow file related syscalls because of a badly implemented security layer](https://discuss.rubyonrails.org/t/why-is-rails-boot-so-slow-on-macos/74021).

```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.report("realpath #{RUBY_PLATFORM}") { File.realpath(__FILE__) }
end
```

2.7.2 on my OSX machine:

```
Warming up --------------------------------------
realpath x86_64-darwin19
                         4.905k i/100ms
Calculating -------------------------------------
realpath x86_64-darwin19
                         49.256k (± 2.4%) i/s -    250.155k in   5.081611s
```

Same machine but in Docker: `docker run -it -v /tmp:/tmp ruby:2.7.2` (so with a FS mount which is known not to be fast with Docker for mac):

```
Warming up --------------------------------------
realpath x86_64-linux
                        51.788k i/100ms
Calculating -------------------------------------
realpath x86_64-linux
                        484.992k (± 5.9%) i/s -      2.434M in   5.036926s
```

So Linux in a Virtual Machine with a file mount is 10 times faster for `realpath` than native OSX....

fxn:
I think I've found a good compromise in https://github.com/fxn/zeitwerk/commit/03cdede1137b0dbe45a5fd93439bbe4289828292, where we compute the real path of the root directories only. No more realpaths are computed.

I think we can close.

fxn:
@casperisfine I have deleted my benchmark above, because I think it was wrong.

In `require ME`, for an already required `ME` (absolute path), I am not really sure what was the benchmark measuring. I have even reduced `$LOADED_FEATURES` to one element to minimize lookup, and it is still way more slow than string equality. Have spent no more time on this, do you know where goes the bulk of the work?

Also, my assertion that `$LOADED_FEATURES` contains real paths (computed by `require` or `require_relative`) is also generally false. I thought that was the case, and that complicated the solution in my head, but discovered late at night that I was wrong. This was an important misconception I had, and I am happy that working on this uncovered it.

The rules are:

1. If you issue `require absolute_path`, `$LOADED_FEATURES` contains `absolute_path` verbatim, no symlinks are resolved whatsoever. Therefore, since paths set by Zeitwerk are absolute, whatever we choose is going to round-trip.

2. If you issue a `require` call to find something in `$LOAD_PATH`, like `require "date"`, the directory part that comes from `$LOAD_PATH` is real, and the rest, as remains as passed to `require`. For example, if directories `foo` and `bar` are symlinks

To summarize the rules above, if `tmp2` is a symlink to `tmp`, and `bar` is a symlink to `foo`, and we execute this

```ruby
require "/Users/fxn/tmp2/bar/x.rb"
require "bar/x"
require "foo/x"```

from `/Users/fxn/tmp2` as `ruby -I. script.rb`, the file is loaded three times, and `$LOADED_FEATURES` contains:

```ruby
[
 "/Users/fxn/tmp2/bar/x.rb",
 "/Users/fxn/tmp/bar/x.rb",
 "/Users/fxn/tmp/foo/x.rb"
]
```

If instead of "." we add `/Users/fxn/tmp2` to `$LOAD_PATH`, you still get its real path in `$LOADED_FEATURES`.

With that misconception clarified, I saw the change was simple and backwards compatible for all practical purposes.

casperisfine:
I tested against our app. `File.realpath` is indeed basically gone:

```
$ stackprof ~/Downloads/stackprof-shopify-boot-production-cpu\ \(1\).dump --method=File.realpath
File.realpath (<cfunc>:1)
  samples:    46 self (0.1%)  /     46 total (0.1%)
  callers:
      26  (   56.5%)  Bootsnap::LoadPathCache::RealpathCache#find_file
       8  (   17.4%)  Bootsnap::CompileCache::YAML::Patch#load_file
       7  (   15.2%)  Zeitwerk::Loader#push_dir
```

And no issue to report so far.

Huge ❤️

fxn:
Awesome :).

Going back to the benchmark, conceptually, if you provide an absolute path to `require`. Mod details (like running code that checks the path is absolute, for example, whatever), the main thing to do is to perform a lookup in `$LOADED_FEATURES` and if found there, that's it. But benchmarks say those "details" amount to a lot.

I've written this one now:

```ruby
require "benchmark/ips"

ME = File.expand_path(__FILE__, __dir__)
$LOADED_FEATURES.clear
$LOADED_FEATURES << ME

Benchmark.ips do |x|
  x.report("lookup") do |n|
    i = 0
    while i < n
      $LOADED_FEATURES.include?(ME)
      i += 1
    end
  end

  x.report("require") do |n|
    i = 0
    while i < n
      require ME
      i += 1
    end
  end

  x.compare!
end
```

However, there has to be a important amount of additional work going on, because the second one is VERY slow in comparison:

```
% ruby -v foo.rb       
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]
Warming up --------------------------------------
              lookup     3.051M i/100ms
             require    87.544k i/100ms
Calculating -------------------------------------
              lookup     27.962M (± 5.4%) i/s -    140.360M in   5.034818s
             require    942.060k (± 3.8%) i/s -      4.727M in   5.025548s

Comparison:
              lookup: 27961530.5 i/s
             require:   942060.0 i/s - 29.68x  (± 0.00) slower
```

I do not know why.

It is dramatically slower in `truffleruby-dev`:

```
% ruby -v foo.rb                 
truffleruby 21.1.0-dev-972a0521, like ruby 2.7.2, GraalVM CE Native [x86_64-darwin]
Warming up --------------------------------------
              lookup   838.476k i/100ms
             require     1.140k i/100ms
Calculating -------------------------------------
              lookup      8.889M (± 5.7%) i/s -     44.439M in   5.018063s
             require     19.451k (± 6.6%) i/s -     96.900k in   5.005607s

Comparison:
              lookup:  8888858.7 i/s
             require:    19451.3 i/s - 456.98x  (± 0.00) slower
```

BTW, I've checked [RubyGems decorated `require`](https://github.com/rubygems/rubygems/blob/058e858c3b9eededac6974adb7d6ca541e39e2c6/lib/rubygems/core_ext/kernel_require.rb) and I wonder if it should short-circuit absolute paths right at the top. I've run the benchmarks with `--disable-gems` and there is some speed up, but still the order of magnitude of difference is similar.

/cc @eregon.

casperisfine:
> I do not know why.

I definitely plan to dig this up later today.

> BTW, I've checked RubyGems decorated require and I wonder if it should short-circuit absolute paths right at the top.

Definitely. You should send a PR (if not I'll do it). There are other useless stuff in there, like `if RUBYGEMS_ACTIVATION_MONITOR.respond_to?(:mon_owned?)`, I bet it's for older rubies. It could be either removed if no longer necessary, or just checked once during code load.

casperisfine:
Wait, I think there's an issue in your benchmark. I kept an investigation log, so I'll post it for posterity and in case I'd need to resume my digging, but I'll post an updated benchmark in a sec. I think the difference you see is because the benchmark is not totally representative.

### Stopped investigation log

First observation the difference is twice as big on OSX than it is on Linux (inside Docker for Mac):

```
Warming up --------------------------------------
              lookup     2.754M i/100ms
             require     4.156k i/100ms
Calculating -------------------------------------
              lookup     18.198M (± 3.7%) i/s -     90.894M in   5.002794s
             require     41.781k (± 2.0%) i/s -    211.956k in   5.075029s

Comparison:
              lookup: 18198337.5 i/s
             require:    41781.0 i/s - 435.56x  (± 0.00) slower
```

vs

```
Warming up --------------------------------------
              lookup     3.103M i/100ms
             require     7.679k i/100ms
Calculating -------------------------------------
              lookup     21.412M (± 3.9%) i/s -    108.597M in   5.080762s
             require     77.779k (± 1.6%) i/s -    391.629k in   5.036488s

Comparison:
              lookup: 21411662.7 i/s
             require:    77778.8 i/s - 275.29x  (± 0.00) slower
```

So that immediately smells like a file related syscall (`stat(2)` or similar). However a quick `strace` shows that it's not the case:

```ruby
ME = File.expand_path(__FILE__, __dir__)
$LOADED_FEATURES.clear
$LOADED_FEATURES << ME

puts '=' * 40
5.times { require ME }
```

```
writev(1, [{iov_base="================================"..., iov_len=40}, {iov_base="\n", iov_len=1}], 2========================================
) = 41
rt_sigaction(SIGINT, {sa_handler=SIG_IGN, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f1b3c69c730}, {sa_handler=0x7f1b3cac5e80, sa_mask=[], sa_flags=SA_RESTORER|SA_SIGINFO, sa_restorer=0x7f1b3c69c730}, 8) = 0
rt_sigaction(SIGINT, {sa_handler=SIG_DFL, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f1b3c69c730}, {sa_handler=SIG_IGN, sa_mask=[], sa_flags=SA_RESTORER, sa_restorer=0x7f1b3c69c730}, 8) = 0
getpid()                                = 321
getpid()                                = 321
timer_delete(0)                         = 0
munmap(0x7f1b3c0e2000, 397312)          = 0
exit_group(0)                           = ?
+++ exited with 0 +++
```

No syscall to be seen, so likely a total red herring.


Now looking a MRI's source. [`require_internal`](https://github.com/ruby/ruby/blob/31ba81788729944705e9f666b52c5ca4990c311d/load.c#L1026-L1118) is where it happens.

In order:

  - `rb_get_path` and `rb_str_encode_ospath`, long story short they coerce the passed String in various way to make sure its encoding matches the OS file system encoding. I've actually worked around this in Bootsnap, for more detail: https://github.com/Shopify/bootsnap/pull/348
  - Then we jump to [`search_required`](https://github.com/ruby/ruby/blob/31ba81788729944705e9f666b52c5ca4990c311d/load.c#L862-L953), which for the most part tries to figure out the extension of the file. If you explicitly pass `.rb` / `.so`, you skip most of it, and pretty much immediately end up in [`rb_feature_p`](https://github.com/ruby/ruby/blob/31ba81788729944705e9f666b52c5ca4990c311d/load.c#L374-L519) which is responsible for searching in `$LOADED_FEATURES`, and should be more or less `O(1)` as it's mostly two hash lookups.

casperisfine:
Argh, nevermind, I'm the one who borked my benchmark. I'll resume my investigation.

eregon:
What's the reason to `$LOADED_FEATURES.clear` in the benchmark?

I see quite different results when I omit that line:
```
$ chruby 2.7.2
$ ruby bench_require_absolute.rb
Warming up --------------------------------------
              lookup   218.499k i/100ms
             require     8.353k i/100ms
Calculating -------------------------------------
              lookup      2.142M (± 0.4%) i/s -     10.925M in   5.099655s
             require     84.119k (± 1.4%) i/s -    426.003k in   5.065292s

Comparison:
              lookup:  2142324.1 i/s
             require:    84119.1 i/s - 25.47x  (± 0.00) slower

$ chruby 3.0.0
$ ruby bench_require_absolute.rb
Warming up --------------------------------------
              lookup   209.673k i/100ms
             require   135.026k i/100ms
Calculating -------------------------------------
              lookup      2.044M (± 2.5%) i/s -     10.274M in   5.029099s
             require      1.331M (± 1.3%) i/s -      6.751M in   5.073517s

Comparison:
              lookup:  2044313.6 i/s
             require:  1330925.6 i/s - 1.54x  (± 0.00) slower
```

Compare that to numbers in https://github.com/oracle/truffleruby/issues/2312#issuecomment-814047198.

Also note some of my findings there, there seem to be a lot more file accesses by RubyGems 3.1.4.

fxn:
> What's the reason to $LOADED_FEATURES.clear in the benchmark?

It was an attempt at making the "details" show up, with the intention of reducing the time spent in the lookup minimal.

casperisfine:
@eregon thanks for pointing the difference between 2.7.2, I was benching with it but looking at ruby-master code 🤦 . That explains why I couldn't figure where time was wasted.

> What's the reason to $LOADED_FEATURES.clear in the benchmark?

I think @fxn wanted to simulate how code reloading works. But I don't think the `clear` is the problem. I think it's the `<<` later on. Basically it's not a good idea to append paths ourselves, as `$LOADED_FEATURES` is not a real Array but a composite object with two internal st_tables.

```ruby
require "benchmark/ips"

ME = File.expand_path(__FILE__, __dir__)
LOADED = "#{ME}.2.rb"
File.write(LOADED, "p :loaded")

# Making sure it load the file twice works as we expect
require LOADED
$LOADED_FEATURES.delete(LOADED)
require LOADED

3.times { require LOADED } # this should basically noop

Benchmark.ips do |x|
  x.report("lookup") do
    $LOADED_FEATURES.include?(LOADED)
  end

  x.report("resolve") do
    $LOAD_PATH.resolve_feature_path(LOADED)
  end

  x.report("require") do
    require LOADED
  end

  x.compare!
end
```


Ruby 3.0.0
```
$ ruby /tmp/bench-require.rb 
:loaded
:loaded
Warming up --------------------------------------
              lookup    23.361k i/100ms
             resolve     5.143k i/100ms
             require    24.164k i/100ms
Calculating -------------------------------------
              lookup    229.462k (± 3.1%) i/s -      1.168M in   5.095627s
             resolve     50.842k (± 2.4%) i/s -    257.150k in   5.060943s
             require    238.176k (± 1.5%) i/s -      1.208M in   5.073819s

Comparison:
             require:   238176.5 i/s
              lookup:   229462.3 i/s - same-ish: difference falls within error
             resolve:    50841.7 i/s - 4.68x  (± 0.00) slower
```

@fxn can you confirm that my benchmark is representative? Zeitwerk never shift stuff in `$LOADED_FEATURES` itself right? If so it seems like on 3.0 we're fine.

fxn:
> Zeitwerk never shift stuff in $LOADED_FEATURES itself right?

In order to be able to reload, it [`rejects!` autoloaded files](https://github.com/fxn/zeitwerk/blob/03cdede1137b0dbe45a5fd93439bbe4289828292/lib/zeitwerk/loader.rb#L323-L336) when it unloads.

fxn:
Confirm that I am getting also these numbers with 3.0.0, and the super slow ones with 2.7.1 (MRI).

Pushing with `<<` instead of going through `require` does not make a difference.

The thing that totally slows things down is the `clear` call! How interesting!

fxn:
My summary:

* RubyGems is way more efficient in 3.0.0 with regard to this than in 2.7.1.
* TruffleRuby is not really different, the benchmark just showed a different RubyGems version was used.
* The benchmark was in addition artificially influenced by the `clear` call. I do not know why, but that was totally artificial in an attempt to minimize lookup time. It has no practical application for Zeitwerk and we can forget it. Lesson learned, `$LOADED_FEATURES` is not a regular array, do not assume the API is transparent.

fxn:
@casperisfine On a second thought, I have removed the `realpath` calls for the root dirs too, the commit has some remarks to explain why (ba58e94db90d022bf77f580dcf5d6f0bdb2e6a96). So, right now there are 0 calls to `realpath`.

As the commit message say, it could be the case that I revisit this, but by now, that's what's in `master`. Just FYI in case it breaks something.

casperisfine:
Thanks for the heads up. I'm pretty certain it won't break anything on our side.

I've added them to the root path just in case, for e.g. capistrano setups (with symlinked `current` directory), but even there I doubt they are actually necessary.

PR #165, opened 2021-06-13
Update README

solnic:
Add a heading "Setting a custom namespace explicitly" under
"File structure" because it's a distinct feature so it deserves its own
section. This way it will be easier to spot it. I was looking for this
feature specifically in the README and missed it, because I clicked on
"Explicit namespaces" header and it was not there, so I assumed it's not
supported.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
 - [Introduction](#introduction)
 - [Synopsis](#synopsis)
 - [File structure](#file-structure)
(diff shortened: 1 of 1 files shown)


fxn:
Looks good!

The README has "Zeitwerk" as a main header, would prefer not to have it in the TOC (maybe it was automatically generated). Would you mind removing that top-level node?

solnic:
@fxn done! and indeed I had an auto-updating TOC enabled in vscode 😅

fxn:
Thanks! ❤️

PR #167, opened 2021-06-30
Load zeitwerk's `VERSION` constant

searls:
Was just working on https://github.com/testdouble/good-migrations/issues/9 and noticed that `Zeitwerk::VERSION` is not available by default, so it seemed like something that should be loaded with the rest of the gem.

Feel free to take it or leave it!


fxn:
Thanks Justin! Guess I overlooked loading this file.

PR #172, opened 2021-07-22
Convert `Zeitwerk::Loader::Config#tag` from accessor to reader

teoljungberg:
This commit converts the `#tag` `attr_acessor` into an `attr_reader` to
make the library warning free. This is caused by the re-definition of
`Zeitwerk::Loader::Config#tag=` further down in the same file.

As we are defining a custom writer, we can convert the `attr_accessor`
into an `attr_reader` safely.


fxn:
Good catch, thanks!

PR #181, opened 2021-09-06
Fix spelling of Kernel/r

datenreisender:
I assume “Kerner” was a spelling mistake, not a joke or something else I do not understand, and it is supposed to be “Kernel”, right?


fxn:
❤️

PR #183, opened 2021-09-15
Add frozen_literal_magic comment in all files

casperisfine:
I was surprised to notice Zeitwerk was quite high up in boot allocation reports:

```
allocated memory by gem
....
  41.42 MB  zeitwerk-2.5.0.beta3

Allocated String Report
   2.20 MB   54947  ".rb"
             48372  zeitwerk-2.5.0.beta3/lib/zeitwerk/loader/helpers.rb:30
...
 
   1.98 MB   49428  "."
             47301  zeitwerk-2.5.0.beta3/lib/zeitwerk/loader/helpers.rb:40
...
 
 654.84 kB   16371  "/"
             10107  zeitwerk-2.5.0.beta3/lib/zeitwerk/loader/config.rb:252
...

```

It's not a big deal, but this simple change can reduce Zeitwerk allocations by over 10%, so IMHO it's worth it.

Also some files had it, others didn't, so it makes sense for consistency sake.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 source 'https://rubygems.org'
 
 gemspec
--- a/Rakefile
+++ b/Rakefile
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'rake/testtask'
 
 task :default => :test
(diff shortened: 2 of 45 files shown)


fxn:
> IMHO it's worth it.

Absolutely. Thanks Jean as always!

PR #203, opened 2022-01-12
Update the Kernel#require comment about prepend

casperisfine:
Ref: https://github.com/Shopify/bootsnap/pull/389

I wondered the same, and was surprised to see it work. Turns out it was fixed recently.

### DIFF
--- a/lib/zeitwerk/kernel.rb
+++ b/lib/zeitwerk/kernel.rb
@@ -17,7 +17,7 @@ module Kernel
   #
   # We cannot decorate with prepend + super because Kernel has already been
   # included in Object, and changes in ancestors don't get propagated into
-  # already existing ancestor chains.
+  # already existing ancestor chains on Ruby < 3.0.
   alias_method :zeitwerk_original_require, :require
 
   # @sig (String) -> true | false


fxn:
Ooohhh, that's excellent!

```ruby
module M
end

class C
  include M
end

module N
  def x
    :x
  end
end

module M
  include N
end

p C.new.x
```

Now that works!

Then, I might refactor this decoration, I prefer `prepend + super` so that you see the decoration in the ancestor chain.

casperisfine:
> Then, I might refactor this decoration, I prefer `prepend + super` so that you see the decoration in the ancestor chain.

Me too, but 3.0 is a bit tricky, it likely mean feature testing and doing one or the other. I'm tempted to do it in Bootsnap. If you end up doing it please let me know.

fxn:
> 3.0 is a bit tricky

In what sense?

> I'm tempted to do it in Bootsnap. If you end up doing it please let me know.

Right now, RubyGems, Bundler, Bootsnap, and Zeitwerk alias `require`. You just adjusted Bootsnap, and if [this proposal](https://github.com/rubygems/rubygems/issues/5263#issuecomment-1007329307) ends up implemented, the four of them will cooperate quite well, I believe.

That reduces a sense of urgency, and if things work well in practice, I'd need to think if having two decorations is worthwhile. Need to have it written to see it.

Dropping support for Ruby < 3.0 some day would be more clear.

casperisfine:
> > 3.0 is a bit tricky
> 
> In what sense?

I meant it means either dropping 2.7 support or doing the work twice. I absolutely don't mind the later, but it means the gems behaves a bit differently on different version, so we'd have to be careful to see how it works with Rubygems and such.

For context I'd like to do this because I'd want to be able to remove Bootsnap hooks post boot, but it's close to impossible to do correctly with alias method chain.

> if [this proposal](https://github.com/rubygems/rubygems/issues/5263#issuecomment-1007329307) ends up implemented

Oh interesting, subscribed!

fxn:
I was trying this refactor, and one test failed, because aliasing in `Kernel` after my decoration had no effect.

Bottom line, if you alias the original method after `prepend`, the alias is ignored:

```ruby
module M
  def f
    p :original
  end
end

class C
  include M
end

module N
  def f
    p :prepended
    super
  end
end

M.prepend(N)

C.new.f

module M
  alias_method :original_f, :f

  def f
    p :aliased
    original_f
  end
end

C.new.f
```

that prints

```
:prepended
:original
:prepended
:original
```

No `:aliased` printed.

Even more weird, if you comment out the first `C.new.f` and leave only the bottom one, the script raises:

```
`f': super: no superclass method `f' for #<C:0x00000001100fc880> (NoMethodError)
```

PR #219, opened 2022-06-07
Eager load constants in a consistent order

casperisfine:
The yield order of `Dir.each_entry` is directly dependent from the file system implementation. On macOS they will be yielded in alphabetical order, because that's how HFS is implemented.

However on Linux, different file system expose different orders. Most commonly, files are iterated with the last modified first, which across a fleet of server is akin to "random".

This small disrependency can cause order dependent code to fail. Arguably, it's the user code that is faulty and should be fixed (I agree), but even then, a predictable order makes it easier to debug.

For that same reason, since Ruby 3 `Dir[]` is sorted: https://bugs.ruby-lang.org/issues/8709

@fxn what do you think?

cc @jeromegn

### DIFF
--- a/lib/zeitwerk/loader/helpers.rb
+++ b/lib/zeitwerk/loader/helpers.rb
@@ -15,7 +15,7 @@ module Zeitwerk::Loader::Helpers
 
   # @sig (String) { (String, String) -> void } -> void
   def ls(dir)
-    Dir.each_child(dir) do |basename|
+    Dir.children(dir).sort.each do |basename|
       next if hidden?(basename)
 
       abspath = File.join(dir, basename)


casperisfine:
Thanks Xavier!

fxn:
Managed to write at least one test for it https://github.com/fxn/zeitwerk/commit/69fdf9847b6fcd02c79d0faf602f8d8f80fbf5c2.

This change is also helpful for log traces, I think. if someone shares code and logs that reproduce something to be debugged, they will be match yours (even in lazy mode).

casperisfine:
> Managed to write at least one test for it

Ahaha thanks, I actually just threw the change quickly to start the discussion, I didn't expect you to merge it right away 😉

fxn:
@casperisfine absolutely :). I was ready to take it over, so just merged and did the rest for speed :).

PR #224, opened 2022-06-24
Allow loader for gem to specify root file

gmcgibbon:
I tried adding Zeitwerk to dry-monitor in https://github.com/dry-rb/dry-monitor/pull/47 and I noticed that the gem loader doesn't support specifying a root file if the loader isn't created in the root file. This use-case is common for gems that extend functionality of other gems (eg. dry-monitor, sprockets-rails, rubocop-rails). 

This patch would allow the Zeitwerk loader I made to be simplified from:
```
@loader ||= Zeitwerk::Loader.new.tap do |loader|
  root = File.expand_path("..", __dir__)
  loader.tag = "dry-monitor"
  loader.inflector = Zeitwerk::GemInflector.new("#{root}/dry-monitor.rb")
  loader.push_dir(root)
  loader.ignore("#{root}/dry-monitor.rb")
  loader.inflector.inflect "sql" => "SQL"
end
```
to:
```
@loader ||= Zeitwerk::Loader.for_gem(root_file: root_file, warn_on_extra_files: false).tap do |loader|
  loader.inflector.inflect "sql" => "SQL"
end
# ...
def root_file
  File.expand_path("../dry-monitor.rb", __dir__)
end
```

Does this make sense?

### DIFF
--- a/lib/zeitwerk/gem_loader.rb
+++ b/lib/zeitwerk/gem_loader.rb
@@ -23,6 +23,7 @@ module Zeitwerk
       @lib                 = File.dirname(root_file)
       @warn_on_extra_files = warn_on_extra_files
 
+      ignore(root_file) if @tag.include?("-")
       push_dir(@lib)
     end
 
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -306,9 +306,9 @@ module Zeitwerk
       # is private, client code can only rely on the interface.
       #
       # @sig (bool) -> Zeitwerk::GemLoader
-      def for_gem(warn_on_extra_files: true)
+      def for_gem(warn_on_extra_files: true, root_file: nil)
         called_from = caller_locations(1, 1).first.path
-        Registry.loader_for_gem(called_from, warn_on_extra_files: warn_on_extra_files)
+        Registry.loader_for_gem(root_file || called_from, warn_on_extra_files: warn_on_extra_files)
       end
 
       # Broadcasts `eager_load` to all loaders.
--- a/test/lib/zeitwerk/test_for_gem.rb
+++ b/test/lib/zeitwerk/test_for_gem.rb
@@ -245,4 +245,24 @@ class TestForGem < LoaderTest
       assert_includes err, "Zeitwerk::Loader.for_gem(warn_on_extra_files: false)"
     end
   end
+
+  test "allows specification of root file" do
+    files = [["lib/my_gem-extension.rb", "require \"my_gem/extension\""], ["lib/my_gem/extension.rb", <<~EOS]]
+      root_file = File.expand_path("../my_gem-extension.rb", __dir__)
+      loader = Zeitwerk::Loader.for_gem(root_file: root_file, warn_on_extra_files: false)
+      loader.enable_reloading
+      loader.setup
+
+      module MyGem
+        module Extension
+        end
+      end
+    EOS
+    with_my_gem(files, false) do
+      _out, err = capture_io do
+        assert require("my_gem-extension")
+      end
+      assert_empty err
+    end
+  end
 end


lib/zeitwerk/gem_loader.rb
```diff
@@ -23,6 +23,7 @@ def initialize(root_file, warn_on_extra_files:)
       @lib                 = File.dirname(root_file)
       @warn_on_extra_files = warn_on_extra_files
 
+      ignore(root_file) if @tag.include?("-")
```
gmcgibbon, line 26: This is a hack. We could try to detect this better with convention, or perhaps add a `ignore_root_file` config option.

fxn:
ACK! There's has been at least a couple of tickets before involving extension gems with dashes. Let me think about this.

fxn:
Something that could help me think about this would be to have a list with a bunch of gems that follow this convention.

gmcgibbon:
- https://github.com/rails/sprockets-rails
- https://github.com/dry-rb/dry-monitor (and all other dry-rb gems AFAICS)
- https://github.com/rspec/rspec-rails
- https://github.com/halostatue/diff-lcs
- https://github.com/mime-types/ruby-mime-types
- https://github.com/aws/aws-sdk-ruby/tree/version-3/gems/aws-partitions
- https://github.com/janlelis/unicode-display_width
- https://github.com/jfelchner/ruby-progressbar

I'm just sifting through a list of most downloaded gems: https://rubygems.org/search?query=downloads%3A+%3E1000000

It would seem that the some gems with dashes aren't extensions, but deliberately namespaced (eg. my_gem = `MyGem` and my-gem = `My::Gem`).

fxn:
Since this is a well-established convention, I believe `for_gem` has to support it out of the box.

If your root file is `ns1/ns2/my_gem.rb` (where the loader is defined), this happens:

* You get a gem loader
* This loader finds a `lib` directory in the ascendants
* Tags the loader as `ns1-ns2-my_gem`
* Ignores `lib/ns1-ns2-my_gem.rb` if exists
* In consequence, if it exists, no warning is issued for this file
* Does not issue a warning for `lib/ns1`

So, you should be able to instantiate a loader as usual in `dry/monitor.rb`, and it should just work as per the conventions.

The WIP patch is https://github.com/fxn/zeitwerk/commit/2bfa619aff044c0770c55349b62204a29ac64279, in the `extensions` branch. Do you have a way to test it? (Documentation not yet written.)

fxn:
It is still as important as always that the namespace being reopened (`Ns1::Ns2` above) gets defined before the loader is setup.

gmcgibbon:
Testing on dry-monitor, your patch works great! However, I have two concerns:

- The loader cannot be initialized in _any_ file (eg. in `my_gem-ext/my_gem/ext.rb` and not `my_gem-ext/my_gem-ext,rb` or `my_gem/ext/loader.rb`) and this may be problematic for some gems. I've seen cases where the entrypoint file is used, or the file structure includes the extension name (eg. `my_gem-ext/my_gem-ext/file.rb`).
- https://github.com/fxn/zeitwerk/compare/main...extensions#diff-d7c22359f0209e4a307f92bd5e441a98c78d5a4f762af2e3ba922349a6e38824R52 The namespace lookup stopping at a directory named `lib` _could_ backfire. A better way to do this may be to look for the gemspec, or even parse the gemspac to see what the root folder is.

These concerns are both pretty atypical to the standard use-cases, but you could solve the first one by allowing the root file to be specified (because the initializer file may not be the root file).

fxn:
My thoughts:

1. I am not 100% sure you should be able to define the loader anywhere in the gem source tree and from there point to any other root file, anywhere. It's two degrees of freedom whose consequences I need to ponder carefully.
2. Agreed regarding `lib`. You might be extending an existing namespace that has `Lib` in the constant path. (The case in which `lib` is a namespace of the gem itself is ruled out by (1)).
3. By now, `for_gem` does not require a gemspec, and this is documented.

fxn:
Another point is that gems are not required to have a `lib` directory. Adding support for other options could be done if there are complains.

fxn:
And another point is that `for_gem` may cover idiomatic patterns only and restrict the options via interface deliberately. If you have something custom, you can always instantiate a regular loader yourself.

gmcgibbon:
> OK, we have a first approximation. I'll give it some more thought before merging.

Sounds good! Since we're going with your patch, I'll close this. Thanks for the reviews and input. I agree we should focus on the idiomatic conventional path that will cover most gems.

fxn:
But right now I don't if I'll release this patch!

fxn:
Oh, that you. At this moment I need to think.

fxn:
I've had had to give it time to find a sweet spot that feels right to me (mod is not written).

The patch I shared has the unacceptable limitation that you cannot have `foo-lib-bar`. No matter if you ship with the solution for that or not, you need to think about how that could look. The answer is: You can optionally pass the gem root directory explicitly. 

However, careful, that could raise the expectation that you could move the definition to `lib/my_gem/ext/zeitwerk.rb` (which is a need because a verbose setup looks off in your entrypoint, more on that later). But then, how do you know if you are in `my_gem-ext_zeitwerk` or not? Do you introduce the concept of an entrypoint? In `foo-bar-baz` is the entrypoint `lib/foo-bar-baz.rb` or is it `lib/foo/bar/baz.rb`? How do we know what to warn about? That adds flexibility in a direction I believe is not going to work well.

So, the trade-offs I have come up with are:

- The README defines what is an entrypoint in conventional gems.
- `for_gem` works without further configuration for conventional gems.
- `for_gem` can only be used in the entrypoint (defined above).
- `for_gem` considers the first ancestor called `lib` to be the gem root directory.
- If your root directory has a different name (the gemspec allows that), you have `gem_root` or something.
- If your extended namespace contains `lib`, you have `gem_root` or something.

I believe that is a sweet spot that makes most conventional gems to have 0 configuration and no warnings.

Now, for non-conventional gems or `lib/my_gem/ext/zeitwerk.rb`, I'll document how to instantiate your own loader, ignore what you have to ignore (`lib/ext`, for example), etc.

I do not abandon the idea of better support for the `lib/my_gem/ext/zeitwerk.rb` usage pattern, but let's give it time.

fxn:
However, if the docs define what the entry point is, and we provide support for an explicit gem root, and we require that whatever is in between is assumed to be an extended namespace and you can't change that, maybe we can make it work with more generality, and still need no config for the common case.

fxn:
The friction point here is that I'd need to ignore `lib/ext` for you, and don't have information to do so. Giving something that seems well resolved to leave a leftover like that is not round design.

PR #230, opened 2022-09-29
Document namespace values for Zeitwerk::Loader::Config#root_dirs

guigs:
Since b5a63c2c00e3f908d78378a5a59659702c175f23 root_dirs does not have `true` as values, but namespace module instead.

### DIFF
--- a/lib/zeitwerk/loader/config.rb
+++ b/lib/zeitwerk/loader/config.rb
@@ -5,18 +5,20 @@ require "securerandom"
 
 module Zeitwerk::Loader::Config
   # Absolute paths of the root directories. Stored in a hash to preserve
-  # order, easily handle duplicates, and also be able to have a fast lookup,
-  # needed for detecting nested paths.
+  # order, easily handle duplicates, to have a fast lookup needed for
+  # detecting nested paths and to store custom namespace as value (default
+  # is Object.
   #
-  #   "/Users/fxn/blog/app/assets"   => true,
-  #   "/Users/fxn/blog/app/channels" => true,
+  #   "/Users/fxn/blog/app/assets"   => Object,
+  #   "/Users/fxn/blog/app/channels" => Object,
+  #   "/Users/fxn/blog/adapters" => ActiveJob::QueueAdapters,
   #   ...
   #
   # This is a private collection maintained by the loader. The public
   # interface for it is `push_dir` and `dirs`.
   #
   # @private
-  # @sig Hash[String, true]
+  # @sig Hash[String, Module]
   attr_reader :root_dirs
 
   # @sig #camelize
--- a/lib/zeitwerk/loader/config.rb
+++ b/lib/zeitwerk/loader/config.rb
@@ -6,12 +6,12 @@ require "securerandom"
 module Zeitwerk::Loader::Config
   # Absolute paths of the root directories. Stored in a hash to preserve
   # order, easily handle duplicates, to have a fast lookup needed for
-  # detecting nested paths and to store custom namespace as value (default
-  # is Object.
+  # detecting nested paths and to store custom namespaces as values (default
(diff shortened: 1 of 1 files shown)


fxn:
Great!

Would like to suggest a few tweaks:

1. While the method `push_dir` has a default, the internal hash `root_dirs` does not. All entries have values. Therefore, I believe we should not have "(default is Object)".
2. "easily handle duplicates" does not have "to", and the new ones would flow better without "to" for coherence.
3. Oxford comma for the "and".

guigs:
Thanks, I made the suggested changes.

By the way, `roots_dir` is considered private and `dirs` the public interface. But `dirs` only return the paths, not the namespaces.

I'm using it because I need to get the namespace for a given root dir. Is there another way? Should `roots_dir` be public?

fxn:
Oh yeah, providing public interface for that was pending. I'll implement it.

No, Zeitwerk does not expose internal collections, otherwise you lose control over them. All mutations have to pass over public methods for you to be able to ensure the consistency of the internal state.

Which is your use case?

guigs:
We are using namespaces to organize a big Rails app into different business domains:

For example:

```
app/domain_a/models
app/domain_a/controllers
app/domain_b/models
etc
```

Then we use `push_dir` to add `app/domain_a/models`, `app/domain_a/controllers`, etc with namespace `DomainA`

We also use gem [Annotate](https://github.com/ctran/annotate_models) to add schema comment to models. But annotate does not know about the namespace and cannot find correct class based on file path.

To fix that we created a patch to lookup the namespace given file path and give it the correct model class (`app/domain_a/models/users` => `DomainA::User`).

For that I implemented a recursive function that navigates to parent dir until a root dir is found:

```ruby
def namespace_for(file)
  rec_namespace_for(File.expand_path(file))
end

def rec_namespace_for(absolute_path)
  Rails.autoloaders.main.root_dirs[absolute_path] || rec_namespace_for(File.dirname(absolute_path))
end
```

So maybe a public API could be similar to that?

fxn:
Cool, I'll think about it. Thanks!

fxn:
@guigs by now, I have added a `namespaces` kwarg to `Zeitwerk::Loader#dirs` and have [documented](https://github.com/fxn/zeitwerk#introspection) the method. Just released version 2.6.1 with this.

I am not so sure about `namespace_for`. I don't expect code using autoloaders to need much programmatic instrospection, prefer to be conservative. This interface is enough for your use case. Might reconsider if time proves me wrong.

guigs:
@fxn Great, thanks for that!

It works for my purpose. One might want to cache the result to avoid multiple allocations, since it creates a new hash for every call. In this sense, it could be a bit easier if there was a method to get the namespace for a specific path.

But I'm not sure this is normal use case. We are using it to fix [the implementation of `get_model_class` method](https://github.com/ctran/annotate_models/blob/ae0b79c0f816b2433a3f00b9a73b3aea1bd3e84b/lib/annotate/annotate_models.rb#L581-L598) from annotate gem.

This is the alternative solution that worked for us:

```ruby
def get_model_class(file)
  loader = Rails.autoloaders.main
  root_dirs = loader.dirs(namespaces: true)
  expanded_file = File.expand_path(file)
  root_dir, namespace = root_dirs.find do |dir, _|
    expanded_file.start_with?(dir)
  end
  _, filepath_relative_to_root_dir = expanded_file.split(root_dir)
  filepath_relative_to_root_dir = filepath_relative_to_root_dir[1..-1].sub(/\.rb$/, '')
  camelize = loader.inflector.camelize(filepath_relative_to_root_dir, nil)
  namespace.const_get(camelize)
end
```

fxn:
Exactly, you'd cache if you need to and you can. The method returns a clone because the internal collection cannot be exposed.

PR #235, opened 2022-10-31
Fix Erroenous Conflicting Directory Error

maths22:
The current implementation ends up clobbering the `dir` variable
after checking the first loader in `Registry.loaders`.  As a result
if you have multiple loaders, the `loader.ignores?(dir)` check
will be guaranteed to fail even if the loader actually ignores
the directory, since there is a stray `/` at the end of the path.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -445,15 +445,15 @@ module Zeitwerk
           next if loader == self
           next if loader.ignores?(dir)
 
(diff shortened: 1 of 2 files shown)


fxn:
You're right, the new `ignores?` uncovered that.

fxn:
I've released 2.6.3 with this fix. Thanks for studying the issue and submitting this patch so quickly!

PR #245, opened 2022-12-28
Fix links in README

teoljungberg:
This PR fixes the link in the README to use the previously added
`a` HTML-links over the direct link to the header via GitHub's URL.

This is consistent with the rest of the README, and there were only a
few outliers:

- `shadowed-files`
- `synopsis`
- `reopening-third-party-namespaces`
- `inflection`


fxn:
❤️

PR #256, opened 2023-01-29
Remove unused call to `register_explicit_namespace`

shioyama:
AFAICT the `autoload_path` here is always a directory, so `ruby?(autoload_path)` is always `false`, hence we can remove this. Tests pass with the removal.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -363,12 +363,10 @@ module Zeitwerk
     # @sig (Module, Symbol, String) -> void
     def autoload_subdir(parent, cname, subdir)
       if autoload_path = autoload_path_set_by_me_for?(parent, cname)
-        cpath = cpath(parent, cname)
-        register_explicit_namespace(cpath) if ruby?(autoload_path)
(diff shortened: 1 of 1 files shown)


fxn:
Not really!

That logic is triggered when you have this situation:

```
app/controllers/foo.rb
app/models/foo/x.rb
```

If `app/controllers` is first in the autoload paths, scanning finds `app/controllers/foo.rb` first, and with the information it has at the moment, that is a regular file for which you have to set an autoload.

Later on, it finds the directory `app/models/foo`. This additional data tells the loader `hotel.rb` was actually defining an explicit namespace, and adjust the internal metadata.

However, your patch made me realize test coverage for this was missing, and also a code comment. I have done this in https://github.com/fxn/zeitwerk/commit/3704206155166a119d5c134484cefd4303bc4097.

BTW, I see the test suite prints a few warnings if running with Ruby 3.2.0, I'll investigate them, the suite should run clean and silent.

shioyama:
Makes sense, thanks! Glad I at least increased test coverage :slightly_smiling_face:

PR #253, opened 2023-01-16
Add minitest-proveit to force assertions

teoljungberg:
This PR adds minitest-proveit to force that all tests executes some kind
of assertion, and if they do not - they will fail automatically.

PR #257, opened 2023-02-07
Replace Rake's testtask with Minitest's

teoljungberg:
This is available since Minitest v5.16.0 and the following commit:

https://github.com/minitest/minitest/commit/73692f9202fbe8dcfe3069708beff155cdffa9bc

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem "rake"
-gem "minitest"
+gem "minitest", "~> 5.16"
 gem "minitest-focus"
 gem "minitest-proveit"
 gem "minitest-reporters"
--- a/Rakefile
+++ b/Rakefile
@@ -1,11 +1,7 @@
 # frozen_string_literal: true
 
-require 'rake/testtask'
+require "minitest/test_task"
 
-task :default => :test
+Minitest::TestTask.create
 
-Rake::TestTask.new do |t|
-  t.test_files = Dir.glob('test/lib/**/test_*.rb')
-  t.libs << "test"
-  t.warning = true
-end
+task default: :test


teoljungberg:
I could add a version lock to the gemfile to denote that we have an expected version of minitest. But since there are no other version constraints I wanted to ask first.

casperisfine:
What is supposed to be the benefit?

teoljungberg:
@casperisfine 

> What is supposed to be the benefit?

Minitest's test task has an easier time passing in options to minitest itself, compared to rake's task.

It also generates more tasks to get the slowests tests etc, if that is of interest in a future refactoring.

That being said - they have the same output which is running tests. I don't mind if we decide to ditch this in favor of rake's test task.

teoljungberg:
As for the "we are not running bundler on CI" - that means we won't get access to tools such as #253 on CI.

But that's a separate issue that I can push a PR for.

teoljungberg:
It looks like bundler is run on CI, as per:

https://github.com/fxn/zeitwerk/pull/258

But then the cache on CI needs to be bumped due to this upgrade change and we do not have that new version of minitest on the CI runs.

teoljungberg:
It's not, minitest 5.16 (where the new test task is introduced) does not support ruby 2.5 as per the CI configuration.

```
conflicting requirements for the Ruby version:
    In Gemfile:
      minitest (~> 5.16) was resolved to 5.17.0, which depends on
        Ruby (< 4.0, >= 2.6)
  
      zeitwerk was resolved to 2.6.6, which depends on
        Ruby (>= 2.5)
  
    Current Ruby version:
      Ruby (= 2.5.9)
```

Therefore with the current Ruby 2.5 and above CI policy - this is not a doable addition.

fxn:
Thanks anyway for looking into this!

PR #266, opened 2023-05-12
Remove unrechable code

exoego:
It appears that `EOS` is not defined nor not reachable.


fxn:
OMG, that has been there [for more than 4 years!](https://github.com/fxn/zeitwerk/commit/f8b5405bd5e3555e5eeaa11ae551c6ed2c1aff5d) Awesome finding!

PR #279, opened 2024-01-04
Fix small typo & clarify some grammar in the docs

stevenharman:
These are just some small language/grammar tweaks that tripped me up while reading through the `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,7 +580,7 @@ root_dir2/my_app/routes
 root_dir3/my_app/routes
 ```
 
-where `root_directory{1,2,3}` are root directories, eager loading `MyApp::Routes` will eager load the contents of the three corresponding directories.
+where `root_dir{1,2,3}` are root directories, eager loading `MyApp::Routes` will eager load the contents of the three corresponding directories.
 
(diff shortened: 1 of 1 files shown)


fxn:
Hey, thanks for proposing this. All good except for two:

> That's a technique used to enable descending into subdirectories on demand

vs the original

> That's a technique used to be able to descend into subdirectories on demand

This is not something you enable/disable wih a toggler, the technique makes lazy descents _possible_. I am not a native speaker, but "being able to" seems to cature the meaning better to me.

The other one is the addition of the article in

> Zeitwerk ignores automatically any file or directory whose name starts with a dot, and any files that do not have an extension ".rb".

The file name "foo.rb" has extension ".rb", not an extension ".rb".

stevenharman:
👋 Hello @fxn, and thank you for the fast feedback! I've force-pushed some further tweaks based on what you said. I think these feel a bit more natural to a native speaker, while also capturing your intent. Happy to also drop these entirely, if you'd prefer.

Thank you, again, for all that you've put into Zeitwerk (and the community at large). ❤️

fxn:
@stevenharman awesome, that revision reads perfect. Thanks a lot for taking the time to edit this, idiomatic English is something I value and cannot produce myself. ❤️

PR #297, opened 2024-07-03
Use require instead of require_relative

JuanVqz:
I wonder if it's "better" or more recommended using require over
require_relative, I think it is, so, suggesting the change, happy to
hear what is best.


fxn:
`require` means "look this relative path up in $LOAD_PATH and load it`. It is a file lookup in a system-wide collection of directories.

`require_relative` means "load this file right here". It is more concise and intentional (and more performant).

PR #301, opened 2024-10-11
Prefer `Object.equal?` over unknown module `equal?` method

byroot:
This is totally nitpicky, but it's easier to assume `Object.equal?` is pristine, while for a random user defined module, it could have been changed.


fxn:
Good point.

PR #302, opened 2024-10-12
Micro-optimization for explicit namespaces

fxn:
Explicit namespaces have an internal hash that maps the cpaths of known explicit namespaces to be loaded to the loader that manages them.

Every added constant needs to verify if there is a loader for the given pair (mod, cname), and that runs for all constants added, globally. So it is a hot spot worth of micro-optimizations.

In particular, for top-level constants (which are common and go through that callback during boot) nowadays we invoke `Symbol#name` and then perform a string lookup. Technically, we can save `Symbol#name` and have a symbol lookup instead, which is faster.

Registering performs an additional comparison now, but registering is more rare than const added.

The hypothesis would be that this may speed boot time up a little bit, and that eager loading would be more or less the same.

This complicates the code a bit, and the hash has mixed keys, which is rare. It would be worthwhile if we see a measurable improvement, otherwise the simpler version would stay.

@byroot what do you think?

### DIFF
--- a/lib/zeitwerk/cref.rb
+++ b/lib/zeitwerk/cref.rb
@@ -13,6 +13,9 @@
 class Zeitwerk::Cref
   include Zeitwerk::RealModName
 
+  # @sig Module
+  attr_reader :mod
+
   # @sig Symbol
   attr_reader :cname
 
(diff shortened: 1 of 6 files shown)


fxn:
In Gusto's monolith I am seeing 1%–2% faster boot times.

byroot:
> what do you think?

I think it's smart. The code is of course a little bit more convoluted, but it really isn't that bad.

fxn:
Cool, let's merge!

PR #304, opened 2024-10-12
Speedup the CI on Windows by skipping the toolchain as it's not needed

eregon:
Compare [before](https://github.com/fxn/zeitwerk/actions/runs/11299692884/usage) and [after](https://github.com/eregon/zeitwerk/actions/runs/11305055840/usage)


fxn:
Awesome!

PR #305, opened 2024-10-18
Add TruffleRuby in CI

eregon:
Now all tests pass, except https://github.com/oracle/truffleruby/issues/2431 for well-known reasons.

I would like to add TruffleRuby in CI so it's easy to notice any potential issue going forward.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,11 @@ jobs:
         ruby-version:
           - "3.2"
           - "3.3"
+          - "truffleruby-head"
           - "head"
+        exclude:
+          - { os: "windows-latest", ruby-version: "truffleruby-head" }
+
     runs-on: ${{ matrix.os }}
     steps:
       - uses: "actions/checkout@v4"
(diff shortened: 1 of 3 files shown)


fxn:
Awesome, thanks @eregon!

.github/workflows/ci.yml
```diff
@@ -16,7 +16,11 @@ jobs:
         ruby-version:
           - "3.2"
           - "3.3"
+          - "truffleruby-head"
```
eregon, line 19: We could now add or switch to truffleruby-24.1.2 in CI, e.g. with `"truffleruby"` (latest release), WDYT @fxn ?

eregon: Right, https://github.com/ruby/setup-ruby/actions/runs/12989568868 was 33min ago and your PR 34min ago so probably just a bit too fast, could you rerun the CI on that PR?

fxn: Greener! There's just [one failure on Windows](https://github.com/fxn/zeitwerk/actions/runs/12989534412/job/36226275463), error says is is unsupported. Should we remove that pair from the matrix?

eregon: Yes, the error is accurate, so that combination should be excluded

fxn: Merged, thanks @eregon!

PR #308, opened 2024-12-16
Manually defining inception namespaces to skip `autoload?` call

gtrias:
In our journey of improving our Rails application bootstrap we have detected that calling the method `Module.autoload?` for each `autoload` registration is very costly.
In our case it takes >3s. (see attached flamegraphs)

image

As the comment in the implementation points out, this call is there to support an edge case on how certain gems autoload themselves. [More info here](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/registry.rb#L28) However this approach produces [a call](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/loader.rb#L540) to `Module.autoload?` for every single constant in our huge project. As it can be seen in the attached flamegraph. This is taking for our backend more than 3 seconds just to support few gems with this special case.

In this PR we propose to enable a mechanism on which any Rails application can manually define the namespaces to be incepted and avoid the costly call for the rest of the constants. This resulted in a boot reduction of more than 3 seconds in our case. We only had to define 3 namespaces to be incepted in order to make our backend run properly.

This is the resulting flamegraph:

image

As a trade-off I understand that this will cause confusion to developers including gems needing the inception. So this PR proposes to keep backwards compatibility if you don't manually set `manual_incepted_namespaces`.

TODO:
- [ ] Add meaningful tests

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -536,9 +536,20 @@ module Zeitwerk
       autoloads[abspath] = cref
       Registry.register_autoload(self, abspath)
 
-      # See why in the documentation of Zeitwerk::Registry.inceptions.
-      unless cref.autoload?
-        Registry.register_inception(cref.path, abspath, self)
+      # If manual incepted namespaces are present we skip the default behavior
+      # and we will only incept the manually defined namespaces for improved
+      # performance
+      if manual_incepted_namespaces.any?
+        if manual_incepted_namespaces.include?(cref.cname.name)
+          Registry.register_inception(cref.path, abspath, self)
+        end
+      else
+        # This operation can impact performance in big codebases since it will
+        # be evaluated for every single auoloaded constant.
+        # See why in the documentation of Zeitwerk::Registry.inceptions.
+        unless cref.autoload?
+          Registry.register_inception(cref.path, abspath, self)
+        end
       end
     end
 
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -539,15 +539,15 @@ module Zeitwerk
       # If manual incepted namespaces are present we skip the default behavior
       # and we will only incept the manually defined namespaces for improved
       # performance
-      if manual_incepted_namespaces.any?
-        if manual_incepted_namespaces.include?(cref.cname.name)
-          Registry.register_inception(cref.path, abspath, self)
-        end
-      else
+      if manual_incepted_namespaces.empty?
         # This operation can impact performance in big codebases since it will
         # be evaluated for every single auoloaded constant.
         # See why in the documentation of Zeitwerk::Registry.inceptions.
         unless cref.autoload?
+          Registry.register_inception(cref.path, abspath, self)
+        end
+      else
+        if manual_incepted_namespaces.include?(cref.cname.name)
           Registry.register_inception(cref.path, abspath, self)
         end
       end
--- a/lib/zeitwerk/loader/config.rb
+++ b/lib/zeitwerk/loader/config.rb
@@ -13,6 +13,9 @@ module Zeitwerk::Loader::Config
   # @sig #call | #debug | nil
   attr_accessor :logger
 
+  # @sig nil
+  attr_accessor :manual_incepted_namespaces
+
   # Absolute paths of the root directories, mapped to their respective root namespaces:
   #
   #   "/Users/fxn/blog/app/channels" => Object,
@@ -84,20 +87,21 @@ module Zeitwerk::Loader::Config
   private :on_unload_callbacks
 
   def initialize
-    @inflector              = Zeitwerk::Inflector.new
-    @logger                 = self.class.default_logger
-    @tag                    = SecureRandom.hex(3)
-    @initialized_at         = Time.now
-    @roots                  = {}
-    @ignored_glob_patterns  = Set.new
-    @ignored_paths          = Set.new
-    @collapse_glob_patterns = Set.new
-    @collapse_dirs          = Set.new
-    @eager_load_exclusions  = Set.new
-    @reloading_enabled      = false
-    @on_setup_callbacks     = []
-    @on_load_callbacks      = {}
-    @on_unload_callbacks    = {}
+    @inflector                  = Zeitwerk::Inflector.new
+    @logger                     = self.class.default_logger
+    @manual_incepted_namespaces = []
+    @tag                        = SecureRandom.hex(3)
+    @initialized_at             = Time.now
+    @roots                      = {}
+    @ignored_glob_patterns      = Set.new
+    @ignored_paths              = Set.new
+    @collapse_glob_patterns     = Set.new
+    @collapse_dirs              = Set.new
+    @eager_load_exclusions      = Set.new
+    @reloading_enabled          = false
+    @on_setup_callbacks         = []
+    @on_load_callbacks          = {}
+    @on_unload_callbacks        = {}
   end
 
   # Pushes `path` to the list of root directories.
--- a/lib/zeitwerk/loader/config.rb
+++ b/lib/zeitwerk/loader/config.rb
@@ -13,7 +13,7 @@ module Zeitwerk::Loader::Config
   # @sig #call | #debug | nil
   attr_accessor :logger
 
-  # @sig nil
+  # @sig Set[String]
   attr_accessor :manual_incepted_namespaces
 
   # Absolute paths of the root directories, mapped to their respective root namespaces:
@@ -89,7 +89,7 @@ module Zeitwerk::Loader::Config
   def initialize
     @inflector                  = Zeitwerk::Inflector.new
     @logger                     = self.class.default_logger
-    @manual_incepted_namespaces = []
+    @manual_incepted_namespaces = Set.new
     @tag                        = SecureRandom.hex(3)
     @initialized_at             = Time.now
     @roots                      = {}
--- a/README.md
+++ b/README.md
@@ -647,6 +647,16 @@ Both strings and `Pathname` objects are supported as arguments. The method raise
 If you want to eager load a directory, `Zeitwerk::Loader#eager_load_dir` is more efficient than invoking `Zeitwerk::Loader#load_file` on its files.
 
 <a id="markdown-reloading" name="reloading"></a>
+
+#### Optimize performance by manually defining inceptions
+
+In order to avoid the costly `autoload?` call on all the autoloaded definitions you can manually define a list of namespaces to be included as inceptions.
+Use this as your own risk, note that after manually setting `manual_incepted_namespaces`, if you add a Gem which needs the inception you will need to include it in this list manually in order to have it working.
+
+```ruby
+loader.manual_incepted_namespaces = ['ActionCable']
+```
+
 ### Reloading
 
 <a id="markdown-configuration-and-usage" name="configuration-and-usage"></a>


lib/zeitwerk/loader/config.rb
```diff
@@ -84,20 +87,21 @@ module Zeitwerk::Loader::Config
...
-    @collapse_glob_patterns = Set.new
-    @collapse_dirs          = Set.new
-    @eager_load_exclusions  = Set.new
-    @reloading_enabled      = false
-    @on_setup_callbacks     = []
-    @on_load_callbacks      = {}
-    @on_unload_callbacks    = {}
+    @inflector                  = Zeitwerk::Inflector.new
+    @logger                     = self.class.default_logger
+    @manual_incepted_namespaces = []
```
jacobobq, line 92: This should be a set if the main operation is going to be `include?`.

lib/zeitwerk/loader.rb
```diff
@@ -536,9 +536,20 @@ def all_dirs
       autoloads[abspath] = cref
       Registry.register_autoload(self, abspath)
 
-      # See why in the documentation of Zeitwerk::Registry.inceptions.
-      unless cref.autoload?
-        Registry.register_inception(cref.path, abspath, self)
+      # If manual incepted namespaces are present we skip the default behavior
+      # and we will only incept the manually defined namespaces for improved
+      # performance
+      if manual_incepted_namespaces.any?
```
jacobobq, line 542: May I suggest [`empty?`](https://rubyapi.org/o/set#method-i-empty-3F) here as perhaps a more idiomatic way to check if a set's empty?

fxn:
Hi!

Let me first understand the situation. As you probably know, Zeitwerk is lazy in development mode and only visits the top directories when an application boots.

In my machine I can do 8 million `autoload?` calls per second, do you have 24 million top-level constants?

gtrias:
> Hi!
> 
> Let me first understand the situation. As you probably know, Zeitwerk is lazy in development mode and only visits the top directories when an application boots.
> 
> In my machine I can do 8 million `autoload?` calls per second when the argument has an autoload set, do you have 24 million top-level constants?
> 
> (When the argument does not have an autoload set, I can do 44 million calls per second.)

Thanks for replying :)

Not sure how to count the number of top-level constants. But if I count the number of times the method `loader#define_autoload` is called in our bootstrap process it gives me 10.779 which is not a lot considering the numbers your machine is able to run `autoload?` in 1 second...

I used this process to count them:

In my `config/application.rb` I set a custom logger for zeitwerk loader like this:

```ruby
    Rails.autoloaders.main.logger = ->(msg) {
      File.open('log/autoloader.log', 'a') { |f| f.write("#{msg}\n") }
    }
    Rails.autoloaders.once.logger = ->(msg) {
      File.open('log/autoloader_once.log', 'a') { |f| f.write("#{msg}\n") }
    }
```

Then I counted number of "autoload set" like this:

```bash
grep 'autoload set' autoloader.log | wc -l
10779

grep 'autoload set' autoloader_once.log | wc -l
5
```

I assume is not a hardware issue either. I have a decent machine and the performance issue is also in our remote devenvs which are similar to my machine.

For reference I have AMD Ryzen 7 5800X 8-Core Processor with 16 cores and 32GB of RAM.

All the test were performed with Ruby 3.3.5.

Any clue why `autoload?` might be so slow for us? It's quite a lot to spend >3 seconds with this process

jacobobq:
Here's my own trace on an Apple M2 Pro, 16GB of RAM.

Speedscope trace

fxn:
Thanks.

I am a bit skeptical of the premise, because while I have been told Factorial is a big application, Shopify is too, and Gusto is too, they are probably even bigger, and that is not an issue.

As I said, above, for the common case in which `autoload?` returns false, I can do 44 million calls in a second.

This is just a heads up, I need to sit down and come up with actual numbers.

fxn:
Could you please confirm the following?

1. When you see 3s saved, is the application eager loading? "Booting" may or may not eager load depending on the environment, you know.
2. Those saved 3s are absolutely 100% only due to this patch?
3. Which command do you use for booting?
4. If you boot the application and inspect the source location of `Module.autoload?`, does it show anything interesting?

fxn:
Spring my be relevant for (2), if present, besides other potential changes.

gtrias:
> I am a bit skeptical of the premise, because while I have been told Factorial is a big application, Shopify is too, and Gusto > is too, they are probably even bigger, and that is not an issue.

TBH I was also skeptical to find such kind of performance issue before bigger companies such the ones you mention did.
Maybe it's something only affecting us because a weird configuration or contextual ad-hoc thingie. But at this point I'm unable to see what could be causing this behavior. Specially because our process of improving the boot performance consists in disabling almost everything and re-enable one by one checking how it affects the perf and trying to optimize it as much as possible.
In this specific case we had our rails app loading in 1,5 secs aprox without loading almost anything. And then we enabled engines loading and we got to ~8 seconds. At this point was when we detected this `Module#autoload?` issue. 

> Could you please confirm the following?
> 
>     1. When you see 3s saved, is the application eager loading? "Booting" may or may not eager load depending on the environment, you know.
>

We only eager-load in production. Our Rails setup is a multi-engine one (with about 168 heterogeneous size engines).
In development environment, the one we're optimizing, we have the following relevant parameters:

```ruby
  # Needed for Spring
  config.enable_reloading = true

  # Do not eager load code on boot.
  config.eager_load = false

  # Use an evented file watcher to asynchronously detect changes in source code,
  # routes, locales, etc. This feature depends on the listen gem.
  config.file_watcher = ActiveSupport::EventedFileUpdateChecker if ENV.fetch(
    'ENABLE_FILE_WATCHER',
    'true'
  ) == 'true'


  # Auto/eager load paths
  config.autoload_paths += Dir.glob("#{config.root}/lib/generators/*")

  # Then we have this lines... That I will honestly try to get rid of and only load what is strictly needed
  autoloader = Rails.autoloaders.main
  autoloader.collapse(Dir.glob("#{config.root}/lib/**/public"))
  config.autoload_lib(ignore: %w[assets tasks])
```


> 
>     2. Those saved 3s are absolutely 100% only due to this patch?
>

Yes, it's the only thing I change between runs and it clearly affects the total time. 
Also the flamegraph confirms that more than 3 seconds was spent by `Module#autoload?` calls 

>
>     3. Which command do you use for booting?
> 

I use to use `rails runner` to test the bootstrap

```
time STACKPROF=true DISABLE_SPRING=true bin/rails runner "puts 'foo'"
....
foo

real    0m26.982s
user    0m21.116s
sys     0m5.676s

# This disables the optimization and fallbacks to the default behavior
time OPTIMIZE_AUTOLOAD=false STACKPROF=true DISABLE_SPRING=true bin/rails runner "puts 'foo'"
....
foo

real    0m32.443s
user    0m23.911s
sys     0m8.390s
```

>
>     4. If you boot the application and inspect the source location of `Module#autoload?`, does it show anything interesting?

We were specially looking for monkey patching around  `Module#autoload?` but we were not able to find anything relevant. Flamegraph traces don't show anything special, the slow part seems to be happening in Ruby internal C implementation of `Module#autoload?`

I will keep investigating. We applied this forked version of Zeitwerk in our `main` branch so it's already in production and all devenvs doing well so far.
I'm also curious to understand why this whole thing is such slow for us.

gtrias:
BTW, to complement this information. I've just confirmed with the infra team that the change also improved our production Puma boot times by ~5seconds

image

fxn:
@gtrias awesome!

Could you please run these commands with engines loaded two times? One with the patch enabled, and one with the patch disabled:

```
bin/rails runner 'p Rails.autoloaders.main.dirs.size'
bin/rails runner 'p Rails.autoloaders.main.unloadable_cpaths.size'
```

That would give me an idea of the magnitude of the project and I could try to reproduce in a synthetic setup.

fxn:
@gtrias I have added a third command in the comment above.

gtrias:
> `bin/rails runner 'p Rails.autoloaders.main.dirs.size'`

Sure!  The env var `OPTIMIZE_AUTOLOAD` enables/disables the `Module#autoload?` optimization

```ruby
    # Manually defining inception namespaces for improved performance
    # More information here:
    if ENV.fetch('OPTIMIZE_AUTOLOAD', 'true') == 'true'
      manual_incepted_namespaces = %w[ActionCable FQL Changelog]
      Rails.autoloaders.main.manual_incepted_namespaces = manual_incepted_namespaces
      Rails.autoloaders.once.manual_incepted_namespaces = manual_incepted_namespaces
    end
``` 

Here the results of the commands you're asking for

```bash
# Without the optimization

time OPTIMIZE_AUTOLOAD=false DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.dirs.size'

1300

real    0m30.804s
user    0m22.679s
sys     0m7.934s

time OPTIMIZE_AUTOLOAD=false DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.unloadable_cpaths.size'

3656

real    0m30.794s
user    0m22.647s
sys     0m7.959s

time OPTIMIZE_AUTOLOAD=false DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.__autoloads.size'

7044

real    0m30.960s
user    0m23.143s
sys     0m7.628s

# With the optimization

time OPTIMIZE_AUTOLOAD=true DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.dirs.size'

1300

real    0m25.590s
user    0m20.333s
sys     0m5.065s

time OPTIMIZE_AUTOLOAD=true DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.unloadable_cpaths.size'

3656

real    0m25.353s
user    0m19.723s
sys     0m5.419s

time OPTIMIZE_AUTOLOAD=true DISABLE_SPRING=true bin/rails runner 'p Rails.autoloaders.main.__autoloads.size'

7044

real    0m25.452s
user    0m19.933s
sys     0m5.326s
```

fxn:
Oh, maybe relevant to your performance work nonetheless, loading 3K files may have a measurable impact perhaps.

gtrias:
I think I have an idea on why calling `Module#autoload?` might have so poor performance for our project.
Looking a Ruby source code it seems that the [internal implementation ](https://github.com/ruby/ruby/blob/master/variable.c#L3042)of `Module#autoload?` there's an optional `recur` attribute defaulting to `true` that makes this `autoload?` check recursive by the nesting of the constant to be checked. 

From my understanding, having very deep nested namespaces such as A::B::C::D::E::F etc could have a negative impact on running this check. And it wouldn't surprise me that we are specially affected since we're kinda addicted of nested namespacing :sweat_smile: 

Do the inception check need to run `autoload?` in recursive mode? I will try to test if it works by disabling recursive to check if that makes the trick as well.

fxn:
If you look at the patch, it is replacing `cref.autoload?`. The variable `cref` stores an internal [Zeitwerk::Cref](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/cref.rb#L38) object whose API abstracts common operations with constants performed by the library.

The flag you founded tells lookup to follow **the ancestor chain** of the receiver, but since the library has no use case for that, [there is a hard-coded](https://github.com/fxn/zeitwerk/blob/18851735df1e97f8984f8af502588ea407792ce3/lib/zeitwerk/cref.rb#L38) `false`.

gtrias:
> If you look at the patch, it is replacing `cref.autoload?`. The variable `cref` stores an internal [Zeitwerk::Cref](https://github.com/fxn/zeitwerk/blob/main/lib/zeitwerk/cref.rb#L38) object whose API abstracts common operations with constants performed by the library.
> 
> The flag you founded tells lookup to follow **the ancestor chain** of the receiver, but since the library has no use case for that, [there is a hard-coded](https://github.com/fxn/zeitwerk/blob/18851735df1e97f8984f8af502588ea407792ce3/lib/zeitwerk/cref.rb#L38) `false`.

Ouch, you're right. Then my theory is totally discarded :disappointed:

fxn:
I have measured in a big project, it has 1600 autoload paths

Since this is strange, I did it in a super primitive way to know exactly what I am measuring: I `bundle open`ed `zeitwerk` and rewrote `autoload?` like this

```ruby
def autoload?
  value = nil
  $seconds_in_autoload += Benchmark.realtime do
    value = @mod.autoload?(@cname, false)
  end
  value
end
```

with `$seconds_in_autoload` initialized to 0 in `config/boot.rb`.

In development mode, we get 0.5s, and it preloads 2K constants.

If you eager load, this app loads 36K constants, and `autoload?` takes 1.6s.

I will investigate if I can reduce those numbers, but it is worth sharing in the thread that the magnitudes are different.

Could you confirm the 3s with this simple technique?

fxn:
I have added a call counter:

```ruby
def autoload?
  $total_autoload_class += 1
  value = nil
  $seconds_in_autoload += Benchmark.realtime do
    value = @mod.autoload?(@cname, false)
  end
  value
end
```

and booting in development mode says the method is called 30K times.

How does that compare in you project?

Since file contents do not matter, once we share these last experiments, I could come up with a self-contained minimal test.

gtrias:
> I have added a call counter:
> 
> ```ruby
> def autoload?
>   $total_autoload_class += 1
>   value = nil
>   $seconds_in_autoload += Benchmark.realtime do
>     value = @mod.autoload?(@cname, false)
>   end
>   value
> end
> ```
> 
> and booting in development mode says the method is called 30K times.
> 
> How does that compare in you project?
> 
> Since file contents do not matter, once we share these last experiments, I could come up with a self-contained minimal test.

Hi, sorry for the late answer. Here are the results of the benchmark:

With the patched version:

```bash
Seconds in autoload: 0.12011783603702497
Total autoload class: 16263

real    0m27.747s
user    0m21.602s
sys     0m5.938s
```

Without patching the autoload behavior:

```bash
Seconds in autoload: 5.537815711977601
Total autoload class: 27143

real    0m32.401s
user    0m23.665s
sys     0m8.626s
```

fxn:
Thanks @gtrias.

So we have 5.54s vs 0.12s spent in that method.

Note that `autoload?` is still called quite a bit 16,263 vs 27,143, so `autoload?` by itself is not causing that difference. It is the method when called exactly in the spot you are patching that is making a difference.

This is really, really strange.

fxn:
So, before anything, I would like to try to find the root cause of that difference in your benchmarks.

I have written this synthetic test. It generates 40,000 files in 1,600 directories, which are root directories of the loader. On setup, this project is bigger than ours (the levels not visited do not matter, we do not need more depth).

The script monkey-patches `Zeitwerk::Cref` so that we can work without editing the gem.

Output in a M2 Pro is:

```
Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload calls: 77615
Seconds in autoload: 0.01
```

Can you reproduce the dramatic slow down in this project?

No matter if you can or if you cannot, any of the options would help us move forward.

```ruby
require 'tmpdir'
require 'fileutils'
require 'benchmark'
require 'zeitwerk'

NUMBER_OF_ROOT_DIRS = 1600
NUMBER_OF_FILES_TOTAL = 40000
NUMBER_OF_FILES_PER_ROOT_DIR = NUMBER_OF_FILES_TOTAL / NUMBER_OF_ROOT_DIRS

$total_autoload_calls = 0
$seconds_in_autoload = 0

module DecorateZeitwerkCrefAutoloadP
  def autoload?
    $total_autoload_calls += 1
    value = nil
    $seconds_in_autoload += Benchmark.realtime do
      value = super
    end
    value
  end

  Zeitwerk::Cref.prepend(self)
end

Dir.mktmpdir do |dir|
  puts <<~EOS
    Generating #{NUMBER_OF_ROOT_DIRS} root directories \
    with #{NUMBER_OF_FILES_PER_ROOT_DIR} files each, \
    for a total of #{NUMBER_OF_FILES_TOTAL} files.
  EOS

  Dir.chdir(dir) do
    # Generate directories and files.
    root_dirs = []
    NUMBER_OF_ROOT_DIRS.times do |n|
      root_dir = "rd#{n}"
      Dir.mkdir(root_dir)
      NUMBER_OF_FILES_PER_ROOT_DIR.times do |m|
        FileUtils.touch("#{root_dir}/file_#{n}_#{m}.rb")
      end
      root_dirs << root_dir
    end

    loader = Zeitwerk::Loader.new
    root_dirs.each { loader.push_dir(_1) }
    loader.setup
  end
end

puts "Total autoload calls: #{$total_autoload_calls}"
puts "Seconds in autoload: #{$seconds_in_autoload.round(2)}"
```

fxn:
@gtrias let me suggest a different experiment. Run that script with the original gem in two ways:

1. With `ruby`.
2. With `bin/rails runner`.

Since (2) loads the script after the application has been initialized, the monkey-patch will be applied after it, therefore only reporting what matters to us.

If (1) takes 0.01s and (2) takes 3s, we got something.

fxn:
I have edited the script to avoid some name clashes (not important, but it is more clean). Now we get 80,000 calls.

gtrias:
Hey! just came back from holidays.

Here are the results with the original version of Zeitwerk and running it with ruby:

```bash
$ time bundle exec ruby scripts/zeitwerk_perf_test.rb 

Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Seconds in autoload?: 0.04

real    0m3.990s
user    0m2.565s
sys     0m1.399s
```

Same but running it with `rails runner`:

```bash
$ time bin/rails runner scripts/zeitwerk_perf_test.rb

Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Seconds in autoload?: 18.47

real    0m56.625s
user    0m37.492s
sys     0m19.165s
```

Same tests with the proposed "patch":

With ruby:

```bash
$ time bundle exec ruby scripts/zeitwerk_perf_test.rb

Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Seconds in autoload?: 0.04

real    0m4.039s
user    0m2.681s
sys     0m1.331s
```

With rails runner:

```bash
$ time bin/rails runner scripts/zeitwerk_perf_test.rb

Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Seconds in autoload?: 18.62

real    0m52.005s
user    0m34.799s
sys     0m16.770s
```

Let me know if you want me to perform any other test :)

fxn:
OK, definitely something is going on.

In isolation, we get times in the order of 100ths of a second. Same code executed with `bin/rails runner` in your case takes 400x.

I don't think Zeitwerk is the issue, since the method we are measuring is just a one-liner wrapper over `Module#autoload?`. I tried in a large code base and I get a 150x factor too. So I can now reproduce. We have moved forward :).

Let me investigate!

fxn:
I have changed the script to generate 400,000 files (instead of 40,000), and the time in isolation is around the expected 10x, linear.

So, my suspicion is that there might be something in the Rails processes responsible for what we are seeing.

fxn:
I have modified the script to distinguish `autoload?` calls that return `true` vs `false` and we are into something.

In the isolated script, we get 40,000 in each case and the `true` case performs a little bit worse:

```
% ruby foo.rb
Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Total autoload? calls (true): 40000
Total autoload? calls (false): 40000
Seconds in autoload? (true): 0.01
Seconds in autoload? (false): 0.0
calls/s (true): 3879350.73
calls/s (false): 8956561.12
```

but in the context of the big Rails app, with the same number of calls for each case, the `true` case becomes the clear outlier (150x factor):

```
% bin/rails runner foo.rb
Generating 1600 root directories with 25 files each, for a total of 40000 files.
Total autoload? calls: 80000
Total autoload? calls (true): 40000
Total autoload? calls (false): 40000
Seconds in autoload? (true): 1.52
Seconds in autoload? (false): 0.01
calls/s (true): 26341.35
calls/s (false): 5567153.44
```

My hypothesis is that there is something going on in the implementation of `Module#autoload?` itself in CRuby, or related code, structures, lookups, mutexes, ... something.

I'll read the source code (not tonight), but let me /cc @byroot, since he is familiar with it and loves performance puzzles :).

The modified script:

```ruby
require 'tmpdir'
require 'fileutils'
require 'benchmark'
require 'zeitwerk'

NUMBER_OF_ROOT_DIRS = 1600
NUMBER_OF_FILES_TOTAL = 40000
NUMBER_OF_FILES_PER_ROOT_DIR = NUMBER_OF_FILES_TOTAL / NUMBER_OF_ROOT_DIRS

$total_autoload_true_calls = 0
$total_autoload_false_calls = 0
$seconds_in_autoload_true = 0
$seconds_in_autoload_false = 0

module DecorateZeitwerkCrefAutoloadP
  def autoload?
    value = nil
    elapsed = Benchmark.realtime do
      value = super
    end
    if value
      $total_autoload_true_calls += 1
      $seconds_in_autoload_true += elapsed
    else
      $total_autoload_false_calls += 1
      $seconds_in_autoload_false += elapsed
    end
    value
  end

  Zeitwerk::Cref.prepend(self)
end

Dir.mktmpdir do |dir|
  loader = Zeitwerk::Loader.new

  puts <<~EOS
    Generating #{NUMBER_OF_ROOT_DIRS} root directories \
    with #{NUMBER_OF_FILES_PER_ROOT_DIR} files each, \
    for a total of #{NUMBER_OF_FILES_TOTAL} files.
  EOS

  Dir.chdir(dir) do
    NUMBER_OF_ROOT_DIRS.times do |n|
      root_dir = "rd#{n}"
      Dir.mkdir(root_dir)
      loader.push_dir(root_dir)

      NUMBER_OF_FILES_PER_ROOT_DIR.times do |m|
        FileUtils.touch("#{root_dir}/d#{n}_f#{m}.rb")
      end
    end
  end

  loader.setup
end

puts "Total autoload? calls: #{$total_autoload_true_calls + $total_autoload_false_calls}"
puts "Total autoload? calls (true): #{$total_autoload_true_calls}"
puts "Total autoload? calls (false): #{$total_autoload_false_calls}"
puts "Seconds in autoload? (true): #{$seconds_in_autoload_true.round(2)}"
puts "Seconds in autoload? (false): #{$seconds_in_autoload_false.round(2)}"
puts "calls/s (true): #{($total_autoload_true_calls / $seconds_in_autoload_true).round(2)}"
puts "calls/s (false): #{($total_autoload_false_calls / $seconds_in_autoload_false).round(2)}"
```

fxn:
> My hypothesis is that there is something going on in the implementation of Module#autoload? itself in CRuby, or related code, structures, lookups, mutexes, ..., something.

Hmmm, not necessarily, because the script in isolation presents some difference in speed, but a normal one. Even with 400,000 files. With the same number of calls.

Maybe it is that, but the Rails context is also a factor in some way or another that we need to understand. It is in the Rails context that we see those 150x and 400x.

And I am going to bed, I promise 😅.

byroot:
> t is in the Rails context that we see those 150x and 400x.

I assume it's only in big apps? I tried with `lobsters` and no difference there. I don't have access to a big app right now so can't repro.

fxn:
@byroot for now, what we can say is that we've only seen this in big apps.

But we do not know yet if being big is the reason. And the fact that the synthetic benchmark shows small times with linear behavior for up to 1600 directories with 400,000 files, does not help concluding the problem might be number of files.

It could be something different, a project dependency or something making this non-linear for some reason or, ..., I don't know for now.

fxn:
I have replaced the call to `cref.autoload?` that returns `true` (the one patched in this PR), with a call to `cref.defined?`, which wraps `const_defined?` with a `false` argument. They should return the same boolean, though the code looks bit less precise to me.

But the slow down is the same.

It could be the case that behind the scenes they do about the same work, would need to double-check. But another bit of data anyway.

fxn:
I have eliminated Zeitwerk from the equation.

In my large app, if I throw

```ruby
require 'benchmark'
autoload :FOO, 'foo.rb'

puts Benchmark.realtime {
  100_000.times do
    Object.const_defined?(:FOO)
  end
}
```

in `config/application.rb` **before** `Bundler.require(*Rails.groups)`, I get 0.01s. If I move it **after** it takes more than 3s.

At that point, not even the application is defined, not not mention any autoloader. Size of the application is, in consequence, not relevant.

Smells to some kind of issue in the project dependencies. But `Gemfile` is +300 LOC, so it needs work, but this is a step forward.

@gtrias does this happen in your project too?

byroot:
Perhaps this can be reproduced in isolation by registering lots of constants on `Object` or lots of `autoload`?

fxn:
Eureka!

I have found the culprit going gem by gem, in my case, the offending dependency is https://rubygems.org/gems/emailage. It is a pretty innocent one, but commenting stuff out I realized is this weird thing at the top of its entry point:

```ruby
$:<<'./lib'
```

That is causing the slow down.

So, the minimal reproduction script is

```ruby
require 'benchmark'

$: << './lib'

autoload :FOO, 'foo.rb'

puts Benchmark.realtime {
  100_000.times do
    Object.const_defined?(:FOO)
  end
}
```

That eliminates Rails.

You can throw that in an empty directory, run it with `ruby`, and it will reproduce.

The next step is to understand why, and it smells something in CRuby.

fxn:
Let me also cc @ioquatix, since he has worked on that code.

Hey! You can ignore the entire thread and read just the previous comment.

You'll see that script takes a disproportionate amount of time, like 3s. No Rails app assumed or anything, just the script executed with `ruby`.

If you comment out the mutation of `$:`, however, the same program takes a more reasonable 0.01s.

The slowdown was detected calling `autoload?` instead of `const_defined?`, but I guess they are doing something common behind the scene.

Also, it is key that the autoload exists. If the constant is defined normally, or does not exist, you get the expected constant time.

fxn:
I have also seen that `lib` is not relevant either, you can push literally `./whatever` and it reproduces.

byroot:
https://share.firefox.dev/4gRLPbs

It's all in `feature_provided_p`, calling mostly `get_expanded_loadpath`.

I need to double check, but I think it is because when it finds a registered autoload, it could potentially be an autoload for something that was already required, and if so it needs to clear it.

fxn:
@byroot that squares, in the sense that is consistent with what we see. At the same time, I'd expect `autoload?` and `const_defined?` to be fast as an API.

Besides, I just set an autoload, as a user, wouldn't you expect `autoload?` to return true ipso facto? If the autoload is triggered, _then_ ordinary `Kernel#require` logic should apply?

byroot:
> I'd expect autoload? and const_defined? to be fast as an API.

I absolutely agree with you. I'm not saying this isn't a bug, or at least that it isn't an undesirable property.

But it still is the result of a desired feature, so it may not be simple to fix (well dunno yet really I haven't put much thoughts into it)

> Besides, I just set an autoload, as a user, wouldn't you expect `autoload?` to return true ipso facto?

Nope!

```ruby
>> File.write("/tmp/a.rb", "")
=> 0
>> require "/tmp/a.rb"
=> true
>> autoload :Bar, "/tmp/a.rb"
=> nil
>> Object.autoload?(:Bar)
=> nil
```

```ruby
>> File.write("/tmp/a.rb", "")
=> 0
>> autoload :Bar, "/tmp/a.rb"
=> nil
>> Object.autoload?(:Bar)
=> "/tmp/a.rb"
>> require "/tmp/a.rb"
=> true
>> Object.autoload?(:Bar)
=> nil
```

fxn:
Yes, yes, I know it does that, and discovered in the sofa midnight, while my family celebrated New Year around in 2019 😂 .

That is the reason I have the `autoload?` call this PR is about. Because Zeitwerk has an inception when you run `setup` in the entrypoint of a gem that is being precisely required, and needs to keep track of that edge case internally.

So, I workaround that, but at first sight it does not feel like an expected API for me.

byroot:
Yup. Just saying we might need to go through a Redmine ticket, depends.

I think we could cache the whole `realpath` thing, but it would need to be invalidated when a `chdir` happens, and I don't know if Ruby is expected `chdir` from C to be supported or not.

Or we could also ask to change that behavior for performance reasons, but that too is a Redmine discussion. I just don't have a strong opinion on the better fix right now as I haven't spent much time thinking about it (and not sure when I will).

There might be some more creative solutions.

fxn:
Agree, it would need at least a Redmine ticket.

My point is that if I say

```ruby
autoload :Foo, 'foo'
```

and the docs of `autoload?` say

> Returns filename to be loaded if name is registered as autoload in the namespace of mod or one of its ancestors.

I expect `autoload?(:Foo)` to return `foo` immediately.

This invalidation logic or `nil` return value is not documented and seems out of place to me.

(Just to elaborate, this would be in such Redmine ticket.)

fxn:
> I think we could cache the whole realpath thing, but it would need to be invalidated when a chdir happens, and I don't know if Ruby is expected chdir from C to be supported or not.

Also if `$LOAD_PATH` is mutated.

ioquatix:
If I'm understanding this correctly, this might be a strong argument for `autoload_relative` as I assume it would bypass the above issues - and I'm assuming adding a relative path to `LOAD_PATH` from a gem is almost certainly a bug.

byroot:
There's just something that hot me. All the `autoload` defined by Zeitwerk have absolute paths. How come the repro needs a relative one?

fxn:
@ioquatix @byroot the problem is not the relative path in the `autoload` call, what slows this down is the relative path in `$LOAD_PATH`.

byroot:
But that's my point. If the path registered in the `autoload` is absolute, I don't think we have to care about what's in `$LOAD_PATH`.

But I'm not 100% on that, I need to dig deeper.

fxn:
```ruby
require 'benchmark'

$LOAD_PATH << 'whatever' # (1)

autoload :FOO, '/tmp/foo'

puts Benchmark.realtime {
  100_000.times do
    Object.autoload?(:FOO)
  end
}
```

if you run that code, and then comment (1) out, you'll see a disproportionate differenece.

byroot:
Yeah, but that may be out way out. Handling both `$LOAD_PATH` and `autoload` entry being relative is tricky, but I think we may be able to skip that overhead when the `autoload` entry is absolute.

byroot:
But again, it's been quite a while since I actively worked on the loading logic, I'm not that fresh on it, I'm probably missing something.

fxn:
@byroot besides the separate discussion about whether that logic makes sense, what you suggest would fix the performance penalty in our use case in practice.

If autoload path is absolute, at least `$LOAD_PATH` does not seem to play any role in that situation.

fxn:
@gtrias let me summarize the status if this.

When an autoload is set for constant `C`, we have seen that `autoload?(:C)` and `const_defined?(:C)` slow down a lot **provided there is a relative path in $LOAD_PATH**. This, at first, seems like a weird connection, because Zeitwerk only uses absolute paths, and `$LOAD_PATH` plays no role there, https://github.com/ruby/ruby/pull/12562 is going to address it. 

However, in a Rails application, adding a relative path to `$LOAD_PATH` should not be necessary, and it is even questionable because people, in general, do not control `chdir` of client code.

If you run `bin/rails runner 'pp $LOAD_PATH`, do you see a relative path?

If you see it, the recommendation is to get rid of it. With that, the overhead in `autoload?` should be mostly gone for current Ruby (and it will be gone, regardless, if https://github.com/ruby/ruby/pull/12562 is applied).

gtrias:
> ```ruby
> require 'benchmark'
> autoload :FOO, 'foo.rb'
> 
> puts Benchmark.realtime {
>   100_000.times do
>     Object.const_defined?(:FOO)
>   end
> }
> ```

Not sure if it's still needed a test from my part, but doing it anyway in case it's relevant:

Running the script before `Bundler.require`:

```
0.09747111699834932
```

Running it after:

```
0.07865373299864586
```

Running it after autoloader configuration: 

```
0.009598845001164591
``` 

I can't reproduce that drastic performance difference you have detected @fxn (unless I do the `$LOAD_PATH << 'whatever'`)

Unfortunately, Our case don't seem to be related with relative paths :sweat_smile: 

```ruby
absolute_paths = $LOAD_PATH.map(&:to_s).filter {|p| p.start_with?('/home') }
absolute_paths.size
=> 2087

relative_paths = $LOAD_PATH.map(&:to_s).reject {|p| p.start_with?('/home') }
relative_paths.size
=> []
```

:thinking: is it possible that we detected another different bug?

fxn:
Thanks @gtrias!

I'll write back with results in that evaluation script we had at the beginning. Just a quick reply to say I don't think pathnames are a problem because Zeitwerk normalizes all user-input paths intro absolute paths as strings, which is what it exclusively uses internally.

byroot:
I opened a ticket on Ruby to potentially backport the bug @fxn found: https://bugs.ruby-lang.org/issues/21032

> Do you think that can make a difference?

It does, just not as big:

```
require 'benchmark'
require 'pathname'

$LOAD_PATH << Pathname.new('/abs-path')

autoload :FOO, '/tmp/foo.rb'

puts Benchmark.realtime {
  500_000.times do
    Object.autoload?(:FOO)
  end
}
```

Takes 600ms on my machine, whereas the same path casted to string is only 50ms. It seems Ruby has to frequently call `to_s` on the `Pathname` which causes some overhead.

Overall, it's probably best if you do like @fxn did to track this down, try to bisect your gemfile by commenting out large pats of it until the problem disapear, etc.

fxn:
Oohhh, right. I had autoload paths in mind in my quick reply, but we are talking about load path entries, actually, which go through what we just saw 👍.

fxn:
I have created https://bugs.ruby-lang.org/issues/21035 about this (to me) unexpected validation logic of `autoload?` and `const_defined?`.

fxn:
I have tested 500,000 `autoload?` calls:

```ruby
require 'benchmark'

autoload :FOO, '/tmp/foo.rb'

puts Benchmark.realtime {
  500_000.times do
    Object.autoload?(:FOO)
  end
}
```

In the case of my large app, the numbers are good provided `$LOAD_PATH` does not have a relative directory:

```
% ruby foo.rb
0.04021999999531545

% bin/rails runner foo.rb
17.698778000019956

% bin/rails runner '$LOAD_PATH.select! { _1.start_with?("/") }; load "foo.rb"'
0.06334600000991486
```

Remember, this application has 1,680 autoload paths, and it is executing `autoload?` on an existing constant about 8,800 times, not 500,000.

This is suggesting to me that `autoload?` on an existing autoload, while slower if we compare with the false case, does not by itself explain 3s in Factorial. And that usage of this API, in principle, should be good.

The challenge now is to understand what's up in Factorial.

gtrias:
> The challenge now is to understand what's up in Factorial.

Let me know if I can test anything else or provide more contextual information to find out it :) 
In the meantime I will keep working on improving our boot and let you know if I find anything relevant to this autoload? topic

fxn:
@gtrias awesome, I'll think about it.

The proposed patch won't probably make it, because one of the key principles of the library is that you don't know who is using it. Whether a dependency loads its code with Zeitwerk, manual `require` calls, or something else, does not belong to their public interface. They document that in order to use the library, you require their entry point, and that is all.

But we'll solve this one way or another :).

fxn:
@gtrias I have rewritten how to handle the "inception" edge case, and the problematic `autoload?` call is gone.

Could you please test the `inception` branch in your project? It would be cool to confirm the overhead is gone, and also that your CI is green (mine is).

fxn:
@gtrias I have worked a lot on that branch (and `main`) this weekend, but think I have to throw it all.

Benchmarks in my big app show no noticeable difference, and it is missing support for an edge case.

gtrias:
> @gtrias I have worked a lot on that branch (and `main`) this weekend, but think I have to throw it all.
> 
> Benchmarks in my big app show no noticeable difference, and it is missing support for an edge case.

Hey! @fxn great, I will test this branch even tho. So, at least, you can know if it improves in our case. Sorry these days I can't dedicate as much time to this topic as I would like. I'll let you know if it improves

fxn:
I have done a second attempt at this today with an important rewrite. However, the performance is the same, and I don't like the compromises I need to take.

At the same time, now that the pathological `./lib` is removed from the big code base I can test, I have measured the time spent in that particular problematic spot: Eager loading, it amounts to around 0.07s over a total of 38s. Even if we make that 3x faster, you won't even notice.

So, give this:

1. The current code uses the autoloading API in a natural way.
2. State is stored where it makes sense.
3. I don't like the alternative approaches I have tried, they require compromises in (1) and (2) that make the implementation worse for my taste.
4. The performance of `Module#autoload?` with an absolute path could even be improved by @byroot's [patch](https://github.com/ruby/ruby/pull/12562).
5. In big code bases, the net effect of that line seems to be negligible.

So, finally, my resolution is going to be that I won't change the current implementation. Always open to consider new evidence, but for now I'm going to close.

Even if the issue is closed, I'd still love to help you understand the root cause in your code base.

PR #311, opened 2025-01-27
Add TruffleRuby stable to CI

fxn:
After addressing https://github.com/oracle/truffleruby/issues/3683, we can start testing against `truffleruby`, and keep an eye on `truffleruby-head` too.

### DIFF
--- a/.github/workflows/keep-an-eye-on-ruby-head.yml
+++ b/.github/workflows/keep-an-eye-on-ruby-head.yml
@@ -14,6 +14,7 @@ jobs:
           - "windows-latest"
         ruby-version:
           - "head"
+          - "truffleruby-head"
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/keep-an-eye-on-ruby-head.yml
```diff
@@ -14,6 +14,7 @@ jobs:
           - "windows-latest"
         ruby-version:
           - "head"
+          - "truffleruby-head"
```
eregon, line 17: I think we'll need an exclude here too for windows + truffleruby-head

fxn: Oh, I had forgotten that the project [has already a daily workflow](https://github.com/fxn/zeitwerk/blob/main/.github/workflows/keep-an-eye-on-truffleruby-head.yml) for `truffleruby-head`! I have removed that line.

Issue #314, opened 2025-05-14
Possible Race Condition for Inflected Constants


aaronmallen:
I am working on breaking out a gem into several smaller gems. In order for this to work I've introduced a plugin system 
that allows the various gems to enhance the functionality of other gems installed in the system. However I think I may
have run into some sort of race condition related to inflections.

We first begin with a simple wrapper for Zeitwerk.

```ruby
# lib/sai/loader.rb

require 'zeitwerk'

module Sai
  class Loader
    def initialize(root_path)
      @loader = Zeitwerk::Loader.new
      @loader.enable_reloading
      @loader.push_dir(File.join(root_path, 'sai'), namespace: Sai)
      @loader.collapse(File.join(root_path, 'sai/errors'))
      @loader.setup
    end

    def add_plugin(root_path)
      @loader.push_dir(File.join(root_path, 'sai'), namespace: Sai)
      @loader.collapse(File.join(root_path, 'sai/errors'))
      @loader.reload
    end

    def collapse(*glob_patterns)
      @loader.collapse(*glob_patterns)
      @loader.reload
    end

    def inflect(...)
      @loader.inflector.inflect(...)
      @loader.reload # <-- note this line.
    end

    def loaded?(constant_name)
      mod = constant_name.delete_prefix('Sai::').split('::').reduce(Sai) do |accumulator, segment|
        break accumulator unless accumulator.autoload?(segment.to_sym).nil?

        accumulator.const_get(segment.to_sym)
      end

      mod&.name == constant_name
    end

    def on_load(module_name, &)
      @loader.on_load(module_name, &)
      @loader.reload
    end
  end
end
```

We then setup Zeitwerk.

```ruby
# lib/sai-core.rb

require_relative 'sai'
require 'sai/loader'

Sai.instance_variable_set(:@loader, Sai::Loader.new(File.dirname(__FILE__)))
Sai.loader.inflect('dsl' => 'DSL') # <-- note this inflection
```

I then create my plugins system:

```ruby
# lib/sai/core/plugin.rb

module Sai
  module Core
    module Plugin
      extend Concurrency

      class << self
        # ...

        def register(plugin_name, root_path = nil, &block)
          return (block ? DSL.process(&block) : nil) if registry.key?(plugin_name)

          raise PluginError, '`root_path` is required when registering a Sai plugin for the first time' unless root_path

          Sai.loader.add_plugin(root_path)

          new_plugins = registry.merge(plugin_name => root_path).freeze
[...]

PR #315, opened 2025-05-16
Allow for root_file customization

aaronmallen:
This allows for specification of where the `for_gem` and `for_gem_extension` root file is located instead of relying souly on `caller_locations` to make that determination. This is useful for allowing custom loader wrapper classes that invoke the loader from places other than the project root.

Further context can be found here:

* https://github.com/fxn/zeitwerk/issues/314
* https://github.com/fxn/zeitwerk/issues/314#issuecomment-2887058756

relates to #314


aaronmallen:
https://github.com/fxn/zeitwerk/issues/314#issuecomment-2887102330

PR #322, opened 2025-12-15
Improve detection of conflicting directories

fxn:
Loaders have to manage disjoint source trees. When a new root directory is configured for a loader, it is checked against all the ones in the registry to see if there is any conflict.

One edge case of this validation is that the paths may apparently overlap, but perhaps the loader is actually ignoring that subtree, so the configuration has to pass. For example, one loader may get `/foo/bar` as root directory, and another loader may get `/foo/bar/baz`.  That seems conflicting, but it isn't if the former loader ignores `/foo/bar/baz`.

In order to do this, we loop, discard ignored paths, and then compare if there may be overlap.

While benchmarking a huge Rails application boot process with +20 autoloaders and ~2000 autoload paths, @ngan noticed checking for ignored stuff showed up. Reason is, if you have ignored directories (which is common in Rails applications), we need to walk up the parent directories of the given one to see if it is a descendant of any of them. Nothing crazy normally, but may add up a bit in large apps with ignored directories.

This patch speeds up the validation.

The key observation is that overlapping configurations are very, very rare. Most of the time, you get loaders taking care of their own gems, and Rails loaders taking care of the application and its engines, and nobody is trying to mess up with paths outside their projects.

So, in practice, it is a better compromise to do the cheap paths comparison and return fast, since most of the time that will already say there is no conflict. Only if we got a potential conflict string-wise, we need to do a last check for ignored paths.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -597,22 +597,24 @@ module Zeitwerk
     #: (String) -> void
     private def raise_if_conflicting_directory(dir)
       MUTEX.synchronize do
-        dir_slash = dir + "/"
-
         Registry.loaders.each do |loader|
           next if loader == self
-          next if loader.__ignores?(dir)
 
           loader.__roots.each_key do |root_dir|
-            next if ignores?(root_dir)
+            # Conflicting directories are rare, optimize for the common case.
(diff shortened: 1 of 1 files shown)


ngan:
Confirming that this change got the process down to `9.92ms`:

image

From `816ms`:

image

PR #335, opened 2026-05-19
Make qualified constant paths work with #on_load

gmcgibbon:
Qualified constant names (`"::A::B"` vs `"A::B"`) should work with Zeitwerk `on_load` callbacks, but right now they currently silently fail. This patch trims prefixed scope operators `::` so that the constant name may be understood by Zeitwerk.

Related to https://github.com/rails/rails/pull/57379, this bug was discovered while implementing an API for class level load hooks in Rails. I'm open to issuing a warning, raising an error, or something else instead. I think the main point here is that we should not silently register these callbacks and never end up calling them. I was curious if you could have a constant in Ruby that starts with `:`, and it seems like you cannot:

```
irb(main):001> class A; end
=> nil
irb(main):002> Object.const_get("::A")
=> A
irb(main):003> Object.const_set("::B", Module.new)
(irb):4:in 'Module#const_set': wrong constant name ::B (NameError)
        from (irb):4:in '<main>'
        ...
irb(main):004> RUBY_VERSION
=> "4.0.1"
```

### DIFF
--- a/lib/zeitwerk/loader/config.rb
+++ b/lib/zeitwerk/loader/config.rb
@@ -278,7 +278,13 @@ module Zeitwerk::Loader::Config
   #
   #: (String?) { (top, String) -> void } -> void ! TypeError
   def on_load(cpath = :ANY, &block)
-    raise TypeError, 'on_load only accepts strings' unless cpath.is_a?(String) || cpath == :ANY
+    case cpath
+    when String
+      cpath = cpath.delete_prefix("::")
+    when :ANY
+    else
+      raise TypeError, 'on_load only accepts strings'
+    end
 
     mutex.synchronize do
       (on_load_callbacks[cpath] ||= []) << block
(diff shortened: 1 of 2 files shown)


fxn:
Hey, this is not a bug, it is deliberate. No constant path in Zeitwerk starts with leading colons, constant paths are assumed to start at top-level, both in input and output.

Because a string is not a constant lookup. When you print the name of a class, it does not have leading colons.

I'll think about the patch!

fxn:
As you said, `const_get` accepts leading colons, so does `const_defined?`. So I think it is coherent that Zeitwerk does the same. I'll add validation too since we are on it.

Thank you @gmcgibbon!

fxn:
@gmcgibbon I just released 2.8.2 including this. Invalid strings raise `NameError` now too, and `on_unload` behaves the same.

Thank you man!

PR #339, opened 2026-08-06
Enable Style/FrozenStringLiteralComment

hmcguire-shopify:
ConstantPathValidator and EagerLoad both currently allocate "::" repeatedly (many thousands of times when eager loading my app).

This commit removes those allocations by enabling frozen_string_literal (using RuboCop so that any new files aren't missed in the future).


fxn:
Thank you!

PR #341, opened 2026-08-30
Raise a controlled error for non-module namespaces

OskarEichler:
When a managed subdirectory maps to an already-defined non-module constant, setup currently leaks `NoMethodError` from the recursive directory walk. Validate the existing value first and raise `Zeitwerk::Error` with the constant path and actual class, matching the controlled namespace-error contract used elsewhere.

Verified with an isolated reproduction on Ruby 4.0.6 and 3.2.11, the complete 528-test / 1,202-assertion suite on both Rubies, RuboCop (80 files), syntax checks and package builds.

### DIFF
--- a/lib/zeitwerk/loader.rb
+++ b/lib/zeitwerk/loader.rb
@@ -573,8 +573,12 @@ module Zeitwerk
       else
         # For whatever reason the constant that corresponds to this namespace has
         # already been defined, we have to recurse.
+        namespace = cref.get
+        unless namespace.is_a?(Module)
+          raise Zeitwerk::Error, "#{cref} is expected to be a namespace, should be a class or module (got #{namespace.class})"
+        end
         log { "the namespace #{cref} already exists, descending into #{subdir}" }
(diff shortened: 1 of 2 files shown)


fxn:
Thank you, could you please add a test for it?

OskarEichler:
Added. The regression predefines the implicit namespace as an Integer, runs setup, and verifies the controlled Zeitwerk::Error and exact constant path/class message. The focused file passes with 15 tests and 24 assertions; targeted RuboCop is clean.

fxn:
Cool, some minor details:

1. Use a regular constant assignment instead of `const_set`.
2. Use the `on_teardown` test helper instead of `ensure`.
3. Use the `remove_const` test helper instead of the `__send__` idiom.

OskarEichler:
Addressed all three details in 8253f3d: the fixture now uses regular constant assignment, registers cleanup with `on_teardown`, and removes the constant through the `remove_const` helper. The focused test remains green at 15 tests / 24 assertions, and targeted RuboCop is clean.

fxn:
Awesome, thank you!

PR #342, opened 2026-08-30
Fix documentation typos

OskarEichler:
Correct eight small persistent typos in contributor instructions, README/changelog prose and one runtime comment. No behavior changes.

The complete 528-test / 1,202-assertion suite passes on Ruby 4.0.6 and 3.2.11, as does RuboCop over all 80 inspected files.


fxn:
Thanks!

PR #344, opened 2026-09-16
Test for compatibility with Ractor's `require` patch

byroot:
Rails main is currently failing, because with all the new Ractor compatibility changes, it triggers a compatibility bug between Zeitwerk and Ractors.

Turns out if you require from a Ractor, you cause a module to be prepended on Kernel, and that breaks Zeitwerk.

```
NoMethodError: super: no superclass method 'require' for main
  rubies/4.0.7/gems/zeitwerk-2.7.3/lib/zeitwerk/core_ext/kernel.rb:34:in 'require'
  rubies/4.0.7/gems/zeitwerk-2.7.3/lib/zeitwerk/core_ext/kernel.rb:34:in 'Kernel#require'
  byroot/rails/railties/lib/rails/application.rb:484:in 'require'
```

It's the usual problem of mixing alias method chain with `prepend`, not sure if there's even a fix for it... If it wasn't for Zeitwerk it would probably break Bootsnap or Rubygems given they also alias method chain `Kernel#require`, but I figured I'd file the issue here so you are aware.


byroot:
Looks like it's  already known upstream: https://bugs.ruby-lang.org/issues/22273

There's probably nothing Zeitwerk could do anyway :/, so probably we should just close.

fxn:
Hey! I see you pushed a workaround in https://github.com/rails/rails/pull/58780 so at least the build is green again.

Let me think about it.

byroot:
I think this is a Ruby problem, and I don't think there's much Zeitwerk can do about it. There's currently an open PR to make it less broken:https://github.com/ruby/ruby/pull/18515

fxn:
@byroot I have read the tickets now, agree about closing.

For the archives: The observable problem from a user point of view is that creating a Ractor before Zeitwerk is loaded breaks Zeitwerk. Top-level autoloads succeed, but anything else, namespaces, callbacks, reloading, etc., won't work.

To reproduce, if we have `Foo = :foo` in _lib/foo.rb_, and run this program:

```ruby
require 'zeitwerk'

Ractor.new {}

loader = Zeitwerk::Loader.new
loader.push_dir('lib')
loader.on_load('Foo') { p :on_load }
loader.setup

p Foo
```

we get the expected output

```ruby
% ruby -W:no-experimental foo.rb
:on_load
:foo
```

But we only need to swap the top lines:

```ruby
Ractor.new {}

require 'zeitwerk'

loader = Zeitwerk::Loader.new
loader.push_dir('lib')
loader.on_load('Foo') { p :on_load }
loader.setup

p Foo
```

to not have the callback invoked as it should:

```
% ruby -W:no-experimental foo.rb
:foo
```

The top-level autoload works, but that is irrelevant, the gem is broken.

The reason why this order matters lies in implementation details of Ractors that are being worked on.

